PR #36, opened 2014-11-06
Gulp: switch to gulp-load-plugins

addyosmani:
In the Gulp community, many of us use [gulp-load-plugins](https://www.npmjs.org/package/gulp-load-plugins) to automatically load any Gulp plugins defined in package.json rather than manually needing to `require` them in.

This patch drops require statements for plugins in package.json and adds support for the plugin. Feel free to close if not particularly useful.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,17 +12,12 @@
 
 'use strict';
 
-var
-  audit = require('gulp-audit'),
-  concat = require('gulp-concat'),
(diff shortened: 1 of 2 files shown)


addyosmani:
@garlicnation I can rebase this if you want to bring it in.

addyosmani:
Let's kill this PR. I can bring it back in at a later date but it doesn't massively change the build setup.

PR #77, opened 2014-11-27
Make HTMLImports polyfill more robust when assigning baseURI property

agrieve:
This came up within the context of Cordova apps (https://github.com/MobileChromeApps/mobile-chrome-apps/issues/450)

On iOS, doc.baseURI is null, and the assignment throws a `TypeError`:

```
TypeError: Attempted to assign to readonly property.
```

I actually don't see when it wouldn't be null since this is a new document. On Chrome, it doesn't throw, but the assignment also doesn't stick.

Instead of calling the setter, this PR creates a new property on the object itself.


mereskin:
https://github.com/webcomponents/webcomponentsjs/issues/190

PR #122, opened 2014-12-12
Fix for a polymer mobile chrome app throwing 'TypeError: Attempted to as...

thomasmarkiewicz:
...sign to readonly property.' in iOS

See here for more info and screenshots: https://github.com/MobileChromeApps/mobile-chrome-apps/issues/457

After this fix I get no more errors and radio buttons work in iOS.

### DIFF
--- a/src/ShadowDOM/wrappers.js
+++ b/src/ShadowDOM/wrappers.js
@@ -156,7 +156,7 @@ window.ShadowDOMPolyfill = {};
   function getSetter(name) {
     return hasEval && isIdentifierName(name) ?
         new Function('v', 'this.__impl4cf1e782hg__.' + name + ' = v') :
-        function(v) { this.__impl4cf1e782hg__[name] = v; };
+        function(v) { if( !(this instanceof Event) ) this.__impl4cf1e782hg__[name] = v; };
   }
 
   function getMethod(name) {


addyosmani:
another ping @thomasmarkiewicz for some further info if you get a chance :)

thomasmarkiewicz:
Just getting back from vacation and getting my head back into this again.  Will follow up shortly with more info...

thomasmarkiewicz:
OK, so I'm using this sample app for testing:

https://github.com/thomasmarkiewicz/polytest2.git

Instructions for building are in the README.  I'm building with both --csp and --strip flags.

After clicking on the Sing Up button, there are two polymer radio buttons to choose Male and Female.  I'm clicking on those in this example which causes the exception.  In this case at the time of the exception:

name === "bubbles"
'this' = 

screen shot 2015-01-07 at 10 30 27 am

On that same page, clicking on the "SIGN UP" button also produces the same exception with same values for 'name' and 'this'

Same results/values when clicking on the "Country" dropdown.

Does this help?

jmesserly:
hey, so I gave it a shot and I can't get an error "Element name could not be inferred." from the vulcanized version. The normal index.html version works though. I couldn't do the README steps that asked me to run "cca" though. Not sure what that command is.

Since it works without vulcanize, this sounds to me like a bug in https://github.com/Polymer/vulcanize

thomasmarkiewicz:
@jmesserly, [cca](https://github.com/MobileChromeApps/mobile-chrome-apps/blob/master/README.md) provides the functionality needed to develop and package Chrome Apps for Mobile from the command line.  I'm trying to develop a mobile chrome app that runs on the iPhone and uses polymer.  Because this is a mobile chrome app, it has content security protection requirements that don't allow incline javascript tags in html, so vulcanize must be run with --csp to put the javascript in separate files or the app won't run at all.

Did you manage to run this app on the iPhone/simulator without cca somehow? If so how - I'd like to debug a bit more with that setup? I think that the iOS browser specifically has something different that causes this exception. 

I can do some work and convert this app to a non-chrome mobile app that runs on top of phonegap directly to eliminate the CSP restrictions and thus the need to vulcanize... that way we'll know if it's vulcanize or not.

jmesserly:
yeah, I just did my normal localhost server (`python -m SimpleHTTPServer` ... there are probably better ways but it usually works) from cca/www and hit localhost:8000 from Safari or iOS simulator. I'm not sure if the app depends on a server but it worked well enough to go through the signup button -> male/female flow. I didn't see any console errors.

But I couldn't get it working with vulcanize (going to localhost:8000/btmzup.html) ... I got the "element name cannot be inferred" error on line 10266 of polymer.js ... looks like a bug in vulcanize? but I'm not sure

jmesserly:
the element it's confused about seems to be:
```js

  (function() {

    var p = {

      eventDelegates: {
        down: 'downAction'
      },

      activeChanged: function() {
        this.super();

        if (this.$.ripple) {
          if (this.active) {
            // FIXME: remove when paper-ripple can have a default 'down' state.
            if (!this.lastEvent) {
              var rect = this.getBoundingClientRect();
              this.lastEvent = {
                x: rect.left + rect.width / 2,
                y: rect.top + rect.height / 2
              }
            }
            this.$.ripple.downAction(this.lastEvent);
          } else {
            this.$.ripple.upAction();
          }
        }

        this.adjustZ();
      },

      disabledChanged: function() {
        this._disabledChanged();
        this.adjustZ();
      },

      recenteringTouchChanged: function() {
        if (this.$.ripple) {
          this.$.ripple.classList.toggle('recenteringTouch', this.recenteringTouch);
        }
      },

      fillChanged: function() {
        if (this.$.ripple) {
          this.$.ripple.classList.toggle('fill', this.fill);
        }
      },

      adjustZ: function() {
        if (!this.$.shadow) {
          return;
        }
        if (this.active) {
          this.$.shadow.setZ(2);
        } else if (this.disabled) {
          this.$.shadow.setZ(0);
        } else {
          this.$.shadow.setZ(1);
        }
      },

      downAction: function(e) {
        this._downAction();

        if (this.hasAttribute('noink')) {
          return;
        }

        this.lastEvent = e;
        if (!this.$.ripple) {
          var ripple = document.createElement('paper-ripple');
          ripple.setAttribute('id', 'ripple');
          ripple.setAttribute('fit', '');
          if (this.recenteringTouch) {
            ripple.classList.add('recenteringTouch');
          }
          if (!this.fill) {
            ripple.classList.add('circle');
          }
          this.$.ripple = ripple;
          this.shadowRoot.insertBefore(ripple, this.shadowRoot.firstChild);
          // No need to forward the event to the ripple because the ripple
          // is triggered in activeChanged
        }
      }

    };

    Polymer.mixin2(p, Polymer.CoreFocusable);
    Polymer(p);

  })();
```

jmesserly:
ah, the bug is here: https://github.com/Polymer/vulcanize/issues/113 ... that's the error I was seeing

thomasmarkiewicz:
OK, good idea.  I'll try the local server route as well as running straight on cordova without vulcanizing.  It could still turn out to be some difference between the mobile Safari and the embedded iOS WebView running in the app.  I believe they're still a bit different at this time - although I don't have a good understanding as to what is different exactly.

thomasmarkiewicz:
Re: Polymer/vulcanize#113, maybe it's related, I'm not sure, but in this case I'm not running vulcanize with --inline.  It'll be a good test to take vulcanize out of the equation completely.  I'll get back with the results.

jmesserly:
eeep, good point about embedded WebView. Since "bubbles" property failed, it could be some Event subtype that for some reason has a non-writable bubbles property. In your screenshot, do you remember the call stack? Someone must've been trying to set "bubbles". Also interesting would be, what does it look like if you expand the "__impl" Event object? That's the real native event object (maybe. Or it's something really strange. :) ).

thomasmarkiewicz:
Callstack:

screen shot 2015-01-07 at 10 37 22 pm

Expanded _impl Event object:

screen shot 2015-01-07 at 10 51 32 pm

jmesserly:
fyi, the "loop" is okay, it's just the wrapper and the Event point at each other :). Cycles are common in the DOM, e.g. childNodes and parentNode.

So it looks like the call stack comes from PointerEvents: https://www.polymer-project.org/platform/pointer-events.html

So that's likely the best place to report this bug.

for what it's worth, I think Polymer is/has moved away from PointerEvents to PointerGestures, but I'm not super familiar. Maybe @azakus can shed some light.

thomasmarkiewicz:
For what it's worth, I don't know for sure but have a hunch that the difference between the cordova version and cca version is that cordova hasEval is true in getSetter but in cca version eval() is not allowed due to CSP so the code takes a different path.  Are there any other test cases you've ran into where webcomponents doesn't have eval available and works?

I was hoping to use polymer to write some iOS packaged chrome apps, but it's turning into too much debug work and no progress.

jmesserly:
yup, folks are using it with CSP in chrome apps. We've been supporting that for a while, AFAIK.

jmesserly:
Although, hmmm. Shadow DOM polyfill code shouldn't even load if you're in Chrome. Chrome has native Shadow DOM support. Not sure if you're seeing that, but I would not expect polyfill to load in Chrome unless you force it on via flags

thomasmarkiewicz:
Yeah, that's my understanding as well that on Android using the crosswalk webview polyfill isn't even needed.  I don't have a problem with Android. Getting that same app running on iOS seems a bit problematic like for example this particular issue.  Or maybe I'm just doing something wrong... are you aware of any sample polymer apps that run on iOS? I need to dig into that a bit more...

jmesserly:
I think all of them run on iOS.  https://www.polymer-project.org/ is built in Polymer, it links some demos at the bottom of the page:

Topeka: https://www.polymer-project.org/apps/topeka/
Calculator: https://www.polymer-project.org/components/paper-calculator/demo.html
Paper elements: https://www.polymer-project.org/components/paper-elements/demo.html#core-toolbar

thomasmarkiewicz:
Yes... but I'd love to see one packaged as a mobile chrome app with the 'cca' tool that runs in iOS embedded WebView ;)

thomasmarkiewicz:
Opened a pull request on Polymer/polymer-gestures that fixes this TypeError exception for me:

https://github.com/Polymer/polymer-gestures/pull/85

thomasmarkiewicz:
Closing this pull request.  Issue has been solved and merge in Polymer/polymer-gestures#85

Issue #115, opened 2014-12-09
Access-Control-Allow-Origin with Polyfill and Amazon S3 (Safari 8 / iOS 8)


tjsavage:
<a href="https://github.com/erwamartin"><img src="https://avatars.githubusercontent.com/u/3288925?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [erwamartin](https://github.com/erwamartin)**
_Wednesday Oct 08, 2014 at 17:50 GMT_
_Originally opened as https://github.com/Polymer/ShadowDOM/issues/516_

---

When I load an external stylesheet from Amazon S3 I get an error in Safari 8 and iOS8 : 

`XMLHttpRequest cannot load https://xg1.s3.amazonaws.com/css/apps/frontend/modules/game/show.min.css?v=2.5.10. Origin http://jsbin.com is not allowed by Access-Control-Allow-Origin. (1, line 0)`

The error occurs because I set shim-shadowdom attribute to my style tag, and Polyfill make an ajax request to get the CSS file. It seems there is no Origin header send in the HTTP request.

In my Amazon S3 dashboard I set a specific configuration to accept CORS :

```
<?xml version="1.0" encoding="UTF-8"?>
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <CORSRule>
       <AllowedOrigin>*</AllowedOrigin>
       <AllowedMethod>GET</AllowedMethod>
       <MaxAgeSeconds>3000</MaxAgeSeconds>
       <ExposeHeader>Location</ExposeHeader>
       <AllowedHeader>*</AllowedHeader>
   </CORSRule>
</CORSConfiguration>
```

When I test from test-cors.org (http://goo.gl/DWzCa6) the resquest succeed but not with Polymer.  
You can test yourself with this example from Sarafri 8 and iOS 8 :  http://jsbin.com/tireg/1

Have you seen this bug?

PR #132, opened 2014-12-16
fix(xhr): add cache busting for safari, webview android and ios

TinOo512:
Because of a ridiculous cache for cross-origin ressources in webkit (cf. https://bugs.webkit.org/show_bug.cgi?id=82672) we need to ensure that we never ask for cached data in those devices.

This must fix this issue (https://github.com/webcomponents/webcomponentsjs/issues/115) in the ShadowDOM repository.

Maybe we should enable this only when we load cross-origin ressources to have a better cache control.

### DIFF
--- a/src/HTMLImports/xhr.js
+++ b/src/HTMLImports/xhr.js
@@ -27,6 +27,18 @@ var xhr = {
       url += '?' + Math.random();
     }
     request.open('GET', url, xhr.async);
+
+    // Safari caches cross-origin resources in a ridiculous way.
+    // We'll work around this annoyance by monkeypatching,
+    // to ensure that we never ask for cached data.
+    // https://bugs.webkit.org/show_bug.cgi?id=82672
+    var isChromium = !!window.chrome; //http://browserhacks.com/#hack-dee2c3ab477a0324b6a2283c434108c8
+    var isFF = !!navigator.userAgent.match(/firefox/i); //http://browserhacks.com/#hack-b3c30c5aee06b9717eda9a5266f0eb07
+    var isIE = window.navigator.msPointerEnabled; //http://browserhacks.com/#hack-2f32c95ac8c021c463de0fdf685acb29
+    if (!isChromium && !isFF && !isIE) {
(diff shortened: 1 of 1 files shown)


garlicnation:
From the bug, it seems like another workaround is to set this header:
```
xhr.setRequestHeader("If-None-Match", "webkit-no-cache");
```

@TinOo512 can you check to see if that resolves this(very weird) caching issue?

addyosmani:
ping @TinOo512 in case you get a chance to reply to the last q on thread.

TinOo512:
@garlicnation @addyosmani Sorry guys completely forget to answer, ...

I'm busy until the end of the week but I can check that at the beginning of next week. 

I'll confirm next week, but I think this will work the same than the `Cache-Control` header because the CDN won't reply with the ETag `webkit-no-cache` for any resources so the browser cached version will always be replaced.

I'll also try to do a Fiddle, like that you will be able to reproduce this (crazy) caching issue.

Just to let you know we are still using this workaround in production even if it's not perfect...

TinOo512:
@garlicnation @addyosmani I tried to reproduce the bug in this codepen : http://codepen.io/TinOo512/pen/JoVYpE but I couldn't succeed. After several test in many webviews none of them failed.

I don't have access to the production server right now, so tomorrow I'll give you two urls of our game, one with my fix and one without. One of them should failed on the iPhone webview for exemple. Like that you'll be able to see the bug.

TinOo512:
@garlicnation @addyosmani I'm not currently working on Polymer, didn’t have much time those last two months, ...

You can try those two examples below on an iPhone's webview (for exemple), one should succeed and the other one fail : 

- This [url](https://polymer.xg1.li/web) is properly loaded. The quiz uses a version of `platform.js` with this fix.

- This other [url](https://polymer.xg1.li/widget), the loader is loading indefinitely because the `XHR` fails for my `css` with the `shim-shadowdom` attributes (at least it's what I understand). The quiz uses the version `0.4.1`  of `platform.js`.

TimvdLippe:
The upstream issue appears to be resolved (https://bugs.webkit.org/show_bug.cgi?id=82672) therefore I am closing this issue.

Issue #134, opened 2014-12-19
Relative path resolving in css


nazar-pc:
```html
<base href="http://cscms.org/uk/">
<link href="/themes/CleverStyle/css/aa.uikit.almost-flat.min.css" rel="stylesheet"
shim-shadowdom>
```
That CSS file contains `url("../fonts/fontawesome-webfont.woff")`, which finally is resolved to `url('http://cscms.org/fonts/fontawesome-webfont.woff')` while should be `url('http://cscms.org/themes/CleverStyle/fonts/fontawesome-webfont.woff')`.

nazar-pc:
Since fix was merged, I'm closing this

PR #135, opened 2014-12-19
Fix for relative path resolving in css

nazar-pc:
Fix for  #134

### DIFF
--- a/src/HTMLImports/path.js
+++ b/src/HTMLImports/path.js
@@ -11,7 +11,6 @@ HTMLImports.addModule(function(scope) {
 
 var CSS_URL_REGEXP = /(url\()([^)]*)(\))/g;
 var CSS_IMPORT_REGEXP = /(@import[\s]+(?!url\())([^;]*)(;)/g;
-var CSS_ABSOLUTE_URL_REGEXP = /^([a-z]+:)?\/\/|data:/i;
 
 // path fixup: style elements in imports must be made relative to the main
 // document. We fixup url's in url() and @import.
@@ -33,8 +32,8 @@ var path = {
   replaceUrls: function(text, urlObj, linkUrl, regexp) {
     return text.replace(regexp, function(m, pre, url, post) {
       var urlPath = url.replace(/["']/g, '');
-      if (!CSS_ABSOLUTE_URL_REGEXP.test(urlPath)) {
-        urlPath = linkUrl.match(/(.*\/)[^\/]+$/)[1] + urlPath;
+      if (linkUrl) {
+        urlPath = (new URL(urlPath, linkUrl)).href;
       }
       urlObj.href = urlPath;
       urlPath = urlObj.href;
--- a/src/HTMLImports/path.js
+++ b/src/HTMLImports/path.js
@@ -11,7 +11,7 @@ HTMLImports.addModule(function(scope) {
 
 var CSS_URL_REGEXP = /(url\()([^)]*)(\))/g;
 var CSS_IMPORT_REGEXP = /(@import[\s]+(?!url\())([^;]*)(;)/g;
-var CSS_ABSOLUTE_URL_REGEXP = /^([a-z]+:)?\/\//i;
+var CSS_ABSOLUTE_URL_REGEXP = /^([a-z]+:)?\/\/|data:/i;
 
 // path fixup: style elements in imports must be made relative to the main
 // document. We fixup url's in url() and @import.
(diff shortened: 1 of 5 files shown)


addyosmani:
If we're improving the path resolution code, it would be useful to add some tests to demonstrate the extra cases it's able to correctly capture.

nazar-pc:
Sure, I'll add them soon.

nazar-pc:
@addyosmani, I've added test but didn't succeed to make it failing without proper fix.

I've tried for so long with `WebComponentsReady`, `HTMLImportsLoaded` and even `DOMNodeRemoved` and `DOMNodeRemovedFromDocument` on `link` node (last two didn't work at all).

The problem is that when for instance `WebComponentsReady` called, original `<link>` still in place which doesn't allow test to fail because browser itself handles that. But if you open `/tests/ShadowCSS/html/style-import-base-tag.html` you can see that after short blink `red` text doesn't actually have red background.

I have such feeling, that this is one more bug in WebComponents.js itself, but I'm not familiar enough with code base to track while is it so (my investigation didn't give any result).

P.S. Firefox 34 & 37 Nightly

nazar-pc:
What about this?
I hope this might be in next release because really makes polyfills unusable in some cases.

src/HTMLImports/path.js
```diff
@@ -11,27 +11,31 @@ HTMLImports.addModule(function(scope) {
 
 var CSS_URL_REGEXP = /(url\()([^)]*)(\))/g;
 var CSS_IMPORT_REGEXP = /(@import[\s]+(?!url\())([^;]*)(;)/g;
+var CSS_ABSOLUTE_URL_REGEXP = /^(http[s]?:)?\/\//i
```
jmesserly: this won't work for schemes other than http(s), right? Is that a concern?
e.g. chrome extensions might use a chrome:// scheme

src/HTMLImports/path.js
```diff
@@ -11,27 +11,31 @@ HTMLImports.addModule(function(scope) {
...
+    var r = this.replaceUrls(cssText, urlObj, linkUrl, CSS_URL_REGEXP);
+    r = this.replaceUrls(r, urlObj, linkUrl, CSS_IMPORT_REGEXP);
     return r;
   },
 
-  replaceUrls: function(text, urlObj, regexp) {
+  replaceUrls: function(text, urlObj, linkUrl, regexp) {
     return text.replace(regexp, function(m, pre, url, post) {
       var urlPath = url.replace(/["']/g, '');
+      if (!CSS_ABSOLUTE_URL_REGEXP.test(urlPath)) {
```
jmesserly: if the URL is parsed using the URL object, it should be easy to check for an absolute path, since at that point the URL is parsed (the parse function at https://github.com/Polymer/URL/blob/master/url.js#L83 ... also it handles more schemes)

jmesserly:
I added some drive-by comments, but probably @sorvell or @addyosmani are best to review this

src/HTMLImports/path.js
nazar-pc: Updated regexp to `/^([a-z]+:)?\/\//i` which will cover any custom scheme

src/HTMLImports/path.js
nazar-pc: Sorry, I didn't get what you meant here.

addyosmani:
I believe John meant that if you use the URL object (shim for the URL standard), it's trivial to check for absolute paths since it already gives you a URL that has been parsed. Have you had a chance to look at using it?

nazar-pc:
Still didn't get what do you mean. Function mentioned by John is long, I haven't even tried to check what it does, regexp looks much simpler for me.

jmesserly:
https://developer.mozilla.org/en-US/docs/Web/API/URL
`new URL(urlPath, linkUrl)` should do correct relative/absolute URL handling without regex.

It's really nice to use libraries because they already handle all of the corner cases correctly. That said, I'm not sure we include the URL polyfill anymore, however IE10+ should have it. @sorvell I recall we used to use URL type but I can't find it used anymore, is that on purpose? If that's the case, I guess the regex munging is the best we can do here.

addyosmani:
> That said, I'm not sure we include the URL polyfill anymore

I wasn't able to find references to the polyfill in our build setup the last time I checked. If we do include it, +1 on directly using it. If not, I think rather than modifying the setup to include another resource regex might indeed be the best way to go.

azakus:
@jmesserly @addyosmani: We're using the URL polyfill in Polymer, but not in webcomponentsjs.
Should be as simple as drafting a pull request to add it to here, and figure out where it should be in the load order.

addyosmani:
@azakus Should the URL polyfill be pulled in using Bower, submoduled, or like the other polyfills, just be checked in?

azakus:
Checking it in should be fine.

nazar-pc:
Replaced regexp by using URL object.
Also test now fails without patch.

jmesserly:
LGTM, thank you!

Issue #12, opened 2014-11-04
document.matches throws a stack-size error when it is added through the platform


PolymerCommunityBot:
<a href="https://github.com/zyklus"><img src="https://avatars.githubusercontent.com/u/320562?v=2" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zyklus](https://github.com/zyklus)**
_Thursday Aug 21, 2014 at 08:21 GMT_
_Originally opened as https://github.com/Polymer/platform/issues/79_

----

`document.matches( '.foo' )`

Will throw a stack size error when the function exists (currently it's only being added for me in Safari 8 in Yosemite, so the bug may be around it being added to #document at all).

garlicnation:
Fixed.

PR #153, opened 2015-01-16
Document.matches shouldn't be defined.

garlicnation:
Solves https://github.com/webcomponents/webcomponentsjs/issues/12.


sorvell:
LGTM

Issue #164, opened 2015-01-23
webcomponentsjs throws an error in Safari, doesn't render elements


jannakha:
webcomponents.js is broken in Safari: a simple demo of paper drop down menu doesn't render (polymer-project.org demos are using old version). Error doesn't say much, but looks like it's the new SVG shadow DOM stuff:

[Error] TypeError: Attempting to configurable attribute of unconfigurable property.
    defineProperty (webcomponents.min.js, line 11)
    j (webcomponents.min.js, line 11)
    H (webcomponents.min.js, line 11)
    (anonymous function) (webcomponents.min.js, line 12)
    global code (webcomponents.min.js, line 12)

dmaslov:
Agree with @jannakha 
Also doesn't work in iOS Safari and iOS Chrome.

miztroh:
Also reported here: https://github.com/Polymer/polymer/issues/1119

addyosmani:
To confirm, this is happening with 0.5.3. For now, please revert to using 0.5.2 until we've rolled a new release to address https://github.com/webcomponents/webcomponentsjs/pull/160 bring merged in.

dmaslov:
@addyosmani thanks for reply!

azakus:
Root cause: Safari has a broken descriptor for `ownerSVGElement`.
````javascript
{get: undefined, set: undefined, configurable: false, enumerable: true}
````

jmesserly:
yeah. see isBrokenSafari check in src/ShadowDOM/wrappers.js

azakus:
We still use the descriptor's `configurable` bit, which is `false` for this property.
Either we have to hack up some path in `registerObject` to have a blacklist, or manually make the wrapper :(

azakus:
Reverted #160 until we can fix it.

arthurevans:
If you're installing from Bower, you can now use 0.5.4, which is identical to 0.5.3 minus the problematic commit.

davej:
Thanks, working again. :+1:

ansorg:
This seems to be again an issue in iOS 9 (Beta 3) on Safari iPhone, webapp using webcomponents version 0.5.5

    TypeError: Attempting to configurable attribute of unconfigurable property.
    definePropertywebcomponents.min.js:10:4150
    jwebcomponents.min.js:10:4150
    Hwebcomponents.min.js:10:4185
    (anonymous function)webcomponents.min.js:11:16300
[...]

PR #160, opened 2015-01-21
Wrap SVGElement.ownerSVGElement

azakus:
Fixes #159

PR #165, opened 2015-01-23
Revert "Wrap SVGElement.ownerSVGElement"

garlicnation:
Reverts webcomponents/webcomponentsjs#160

#160 is the likely cause of #164


addyosmani:
cc @jmesserly

azakus:
LGTM

PR #167, opened 2015-01-24
element.matches() fixes

nazar-pc:
Support for `:host`, `::content` and similar selectors in `element.matches()`

### DIFF
--- a/src/ShadowDOM/querySelector.js
+++ b/src/ShadowDOM/querySelector.js
@@ -49,7 +49,27 @@
   }
 
   function shimSelector(selector) {
-    return String(selector).replace(/\/deep\//g, ' ');
+    return String(selector)
+      // Transform `:host(selector)` to `selector`
+      .replace(
+        /:host\(([^\s]+)\)/g,
+        '$1'
+      )
+      // Transform `selector:host` to `selector`
+      .replace(
+        /([^\s]):host/g,
+        '$1'
+      )
+      // Transform `:host` to `*`
+      .replace(
+        ':host',
+        '*'
+      )
+      // From ShadowCSS, will be replaced by space
+      .replace(
+        /\^|\/shadow\/|\/shadow-deep\/|::shadow|\/deep\/|::content/g,
+        ' '
+      );
   }
 
   function findOne(node, selector) {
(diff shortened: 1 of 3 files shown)


nazar-pc:
Without this patch will fail with `An invalid or illegal string was specified`

addyosmani:
Do you have a test case that fails without this patch in place?

nazar-pc:
Is jsfiddle ok for demonstration (webcomponents.js included from polymer-project.org)?
http://jsfiddle.net/u6vdvd5q/
It will fail to run actually, look into console for `An invalid or illegal string was specified`

Just on any page with WebComponents.js write something like `document.querySelector('whatever').matches(':host *')` and it will break.

However, with patched version it will work, the same example with patched build embedded:
http://jsfiddle.net/u6vdvd5q/1/

addyosmani:
Confirmed via your test case. @garlicnation do we want to land this?

jmesserly:
@addyosmani IMO, it's great to support these. It's kind of strange that querySelector and matches are inconsistent right now.

A few high level comments on the change:

it would be nice to put the logic in https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/querySelector.js so it's all grouped together

also, https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/ShadowRenderer.js and querySelector both rely on el.matches() ... is it intended that they're now calling this polyfilled method? I worry that we're checking for something twice. But maybe this actually fixes some bugs. (apologies, I don't have the querySelector code loaded into my brain anymore, so off hand not sure how they'll interact :) ). Finally, I'd double check that this doesn't hurt performance of qS/qSA since matches is polyfilled now, and it will be computing that polyfilled selector string over and over.

jmesserly:
btw, an easy way to avoid my concerns about interactions would be to change the existing el.matches calls in ShadowDOM code to call originalMatches.

nazar-pc:
Currently `originalMatches` is local variable and not available anywhere outside its scope.

jmesserly:
yeah, no worries about that though, it can be easily saved:

```js
scope.originalMatches = originalMatches;
```

addyosmani:
@nazar-pc Want to go ahead and update your PR to make `originalMatches` available outside its scope? It would be great to land this once @jmesserly's comments have been addressed.

nazar-pc:
Done.
By the way I found that `document.querySelector('div::shadow div')` doesn't work with polyfill, while working natively in Chromium. That is why I've fixed that too and added tests for that.

Both `querySelector` and `ShadowRenderer` use overridden version of `.matches()` because a lot of tests fail with original.

nazar-pc:
Sorry about mixing `querySelector` fix in this pull request, but they are really very close.

jmesserly:
> Sorry about mixing querySelector fix in this pull request, but they are really very close.

no worries! taking a look now

jmesserly:
LGTM. thank you!

PR #175, opened 2015-01-28
Point package.json to the actual main script

shama:
If you point the `main` in the `package.json` to the `webcomponents.js` script, I can require it using browserify or webpack with:

```js
require('webcomponents.js')
```

As oppose to currently needing to do:

```js
require('webcomponents.js/webcomponents.js')
```

Thanks!


addyosmani:
LGTM. Thanks, Kyle!

PR #178, opened 2015-01-28
install createShadowRoot only if present, fixes #174

### LINKED ISSUE #174: Compiled "lite" build without Shadow DOM still defines createShadowRoot
The latest compiled build of the no-Shadow-DOM "lite" version still defines `createShadowRoot`, here is the block: https://github.com/webcomponents/webcomponentsjs/blob/0.5.4/webcomponents-lite.js#L1365-L1370.

This causes downstream code using the presence of the `createShadowRoot` method for feature detection, to fail.

jmesserly:
Pretty simple fix. Since ShadowDOM polyfill must always be loaded first, this should work in both webcomponents-lite.js and webcomponents.js mode

### DIFF
--- a/src/CustomElements/observe.js
+++ b/src/CustomElements/observe.js
@@ -289,13 +289,17 @@ function upgradeDocumentTree(doc) {
 }
 
 
-// ensure that all ShadowRoots watch for CustomElements.
+// Patch `createShadowRoot()` if Shadow DOM is available, otherwise leave
+// undefined to aid feature detection of Shadow DOM.
 var originalCreateShadowRoot = Element.prototype.createShadowRoot;
-Element.prototype.createShadowRoot = function() {
-  var root = originalCreateShadowRoot.call(this);
-  CustomElements.watchShadow(this);
-  return root;
-};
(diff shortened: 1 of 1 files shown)


jmesserly:
fyi, I tested this against Polymer 0.5.x w/ webcomponents.js and 0.8-preview branch w/ webcomponents-lite.js as a sanity check in Firefox/Safari/Chrome. All tests passed.

@addyosmani or @sorvell or @azakus would you mind reviewing?

jmesserly:
or feel free to merge/tweak as desired :)

addyosmani:
Changes LGTM. Tested and verifying as expected.

addyosmani:
@csuwildcat until we cut a new release, you might need to work off master. Could you verify that with this change in place everything works as you would expect?

jmesserly:
thank you @addyosmani :)

btw, another possible workaround would be to use `window.ShadowRoot` in the meantime. Though agreed Element.prototype.createShadowRoot is cleaner to test for

csuwildcat:
@addyosmani will do, I'll report back shortly.

PR #182, opened 2015-01-30
Fix embed tag behavior in firefox when flash has already loaded.

### LINKED ISSUE #181: Error calling installProperties on a flash <embed> object in Firefox
Here's a minimal reproduction of the bug:

```html
    <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
    <script src="http://www.polymer-project.org/components/webcomponentsjs/webcomponents.js"></script>
    <embed src="http://www.youtube.com/v/dQw4w9WgXcQ" type="application/x-shockwave-flash" pluginspage="https://www.macromedia.com/go/getflashplayer">
```

In Firefox, with Flash installed and running when the page initializes, we get

```Error: Error enumerating properties on scriptable plugin object```

This error originates in installProperty, where we call getOwnPropertyNames on the prototype of unwrapped elements. In Firefox with Flash enabled, the prototype obtained by calling getPrototypeOf on a flash <embed> element is `NPObject JS wrapper class { }`. This NP Object will throw the above error when getOwnPropertyNames is called on it.

You can reproduce the specific code that lead to this error by running the following in the FF console:

```javascript
   Object.getOwnPropertyNames(Object.getPrototypeOf(unwrap(document.querySelector('embed'))))
```

garlicnation:
Fixes #181

### DIFF
--- a/src/ShadowDOM/wrappers.js
+++ b/src/ShadowDOM/wrappers.js
@@ -103,6 +103,17 @@ window.ShadowDOMPolyfill = {};
 
   function getWrapperConstructor(node) {
     var nativePrototype = node.__proto__ || Object.getPrototypeOf(node);
+    if (isFirefox) {
+      // HTMLEmbedElements will sometimes be [NS Object wrapper class]
+      // which throws an error when getOwnPropertyNames is called on it.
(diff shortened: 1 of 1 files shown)


jmesserly:
nice! maybe squash changes? otherwise LGTM :)

PR #191, opened 2015-02-09
Fixes exception, thrown in Safari 7 on iPad

### LINKED ISSUE #190: Mobile Safari 7 fails with exception when trying to define `baseURI` property on `doc`
screen shot 2015-02-09 at 3 02 30 pm

Reproducible on an iPad w/ iOs7

mereskin:
fixes #190

### DIFF
--- a/src/HTMLImports/importer.js
+++ b/src/HTMLImports/importer.js
@@ -119,6 +119,10 @@ function isLinkRel(elt, rel) {
   return elt.localName === 'link' && elt.getAttribute('rel') === rel;
 }
 
+function hasBaseURIAccessor(doc) {
+  return !! Object.getOwnPropertyDescriptor(doc, 'baseURI');
+}
+
 function makeDocument(resource, url) {
   // create a new HTML document
   var doc = document.implementation.createHTMLDocument(IMPORT_LINK_TYPE);
@@ -128,7 +132,7 @@ function makeDocument(resource, url) {
   var base = doc.createElement('base');
   base.setAttribute('href', url);
   // add baseURI support to browsers (IE) that lack it.
-  if (!doc.baseURI) {
+  if (!doc.baseURI && !hasBaseURIAccessor(doc)) {
     // Use defineProperty since Safari throws an exception when using assignment.
     Object.defineProperty(doc, 'baseURI', {value:url});
   }


jmesserly:
I tend to think of try+catch as a last resort ... I wonder if we can figure out if it will throw before calling it. Does Object.getOwnPropertyDescriptor(doc, 'baseURI') return something useful? (where `doc` is one of: document, HTMLDocument.prototype, Document.prototype). What about `document.baseURI`?

mereskin:
I'll check tomorrow.

mereskin:
In Safari 7 `baseURI` accessor is already defined on document and is not configurable. I updated conditional and now tests pass in Safari 7.

IE11 has 3 failing tests, but they also fail on master. I don't have IE10 VM handy.

src/HTMLImports/importer.js
```diff
@@ -128,7 +132,7 @@ function makeDocument(resource, url) {
   var base = doc.createElement('base');
   base.setAttribute('href', url);
   // add baseURI support to browsers (IE) that lack it.
-  if (!doc.baseURI) {
+  if (!hasBaseURIAccessor(doc)) {
```
jmesserly: I wonder if we should keep both checks here? In other words:
```js
  if (!doc.baseURI && !Object.hasOwnProperty(doc, 'baseURI')) {
```

(also can use hasOwnProperty instead of getOwnPropertyDescriptor, since we just want a boolean result)

Otherwise I worry this might send correct browsers down the wrong path. baseURI should never be an own-property of `document`... I guess old Safari was broken. According to the spec, it should be on Node.prototype: https://dom.spec.whatwg.org/#interface-node

jmesserly:
added one comment, but overall I like the fix. thanks for tracking that down.

mereskin:
screen shot 2015-02-11 at 11 56 14 am

I amended commit to keep `!doc.baseURI`, but `Object.hasOwnProperty(doc, 'baseURI')` is false, so I'd rather leave it as it is.

jmesserly:
oh wow, good idea to check that. yeah I guess since the accessor is busted, it's possible for getOwn* and hasOwn* to be different. Crazy. Anyway this is great, going to merge. Cheers!

mereskin:
Thank you!
Do you know when is the planned date of a new release?
We happened to use the polyfill in production, and some of our clients await for the fix.

garlicnation:
0.5.5 just came out. I'm continuing to work on test infrastructure/CI part-time, but hopefully during march we'll have continuous delivery of webcomponentsjs so that every merged pull request fires off a release.

mereskin:
Thank you, great news!

PR #196, opened 2015-02-12
Adds the URL polyfill to sources

addyosmani:
This adds a snapshot of the URL polyfill to the sources directory as part of our work evolving #135. Confirmed with Daniel that we would prefer checking this in vs. using Bower to pull from the other repo. 

I'll follow up with a proposal for where this should appear in the load order during builds.


jmesserly:
lgtm

addyosmani:
Thanks @jmesserly! @garlicnation as with #197 will leave up to you to merge as you see fit.

PR #197, opened 2015-02-12
Include URL polyfill references in lite/full builds

addyosmani:
(Please don't merge until #196 lands)

Continuing changes for #135, this change specifies the URL polyfill as a build-time dependency before HTML Imports. Given that the use-case for us landing URL polyfill support was Imports, I believe this is the right place.

Feel free to let me know if not.

### DIFF
--- a/src/WebComponents/build-lite.json
+++ b/src/WebComponents/build-lite.json
@@ -1,5 +1,6 @@
 [
   "build/boot.js",
+  "../URL/URL.js",
   "../HTMLImports/build.json",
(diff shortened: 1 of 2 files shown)


jmesserly:
this lgtm too :)

addyosmani:
@garlicnation as we've landed #196, shall I go ahead and rebase this against master?

garlicnation:
@addyosmani goforit!

addyosmani:
@garlicnation rebased against master. Sorry for the delay!

PR #198, opened 2015-02-12
Add browser support matrix

### LINKED ISSUE #65: Be more specific on browser compatibility
I think it's valuable to include more information about browser support in the README and in the [webcomponents.org/polyfills](http://webcomponents.org/polyfills) page.

Is this list up to date? https://www.polymer-project.org/resources/compatibility.html

addyosmani:
Fixes #65. We've discussed adding a revised browser support matrix to the project for a while. Consider this a first attempt at defining that.

Renders:

screen shot 2015-02-12 at 17 23 16

Some considerations:

* I've attempted to be consistent with our old matrix in saying IE10 support for CE + Imports are flaky due to mutation observers. Is our position here still the same?
* We've intentionally not noted support for IE9 in the past. Developers are however attempting this - do we want to add a note in the README calling out these efforts?
* Should Safari (desktop) join the other vendors in being latest version only, or is 7+ sufficient?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,22 @@ Pre-built (concatenated & minified) versions of the polyfills are maintained in 
 `webcomponents-lite.js` includes all polyfills except for shadow DOM.
 
 
+## Browser Support
+
+Our polyfills are intended to work in the latest versions of evergreen browsers. See below
+for our complete browser support matrix:
+
+|   | IE10 | IE11+ | Chrome* | Firefox* | Safari 7+* | Chrome Android* | Mobile Safari* |
+|---|:---:|:---:|:---:|:---:|:---:|:---:|
+| __Custom Elements__ | ~ | x | x | x | x | x| x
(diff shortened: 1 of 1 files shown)


azakus:
Can we make those "X"s into a more "✓"?

jmesserly:
also would be nice to add details about "flaky"? ... If I recall, that means MutationObserver polyfill? ... if using CustomElements or HTMLImports with ShadowDOM, then they would get not-flaky MutationObserver polyfill that SD includes.

addyosmani:
> Can we make those "X"s into a more "✓"?

Absolutely. I'll update once we've confirmed what our support levels should be.

> also would be nice to add details about "flaky"?

Agree. Details on flakiness are also reliant on which browsers we decide to include support for. Will flesh this out further shortly.

addyosmani:
After confirming the browser support matrix with @tjsavage, I've made the adjustments suggested in the comments and added an additional note to cover browsers outside of our matrix.

Landing, happy to edit as needed!

PR #201, opened 2015-02-13
Fixes #137 - adds LICENSE informtion

### LINKED ISSUE #137: Add LICENCE file
This repo is missing a licence declaration.  I'd like to reuse some of the polyfills in the [polyfill service](https://cdn.polyfill.io) but won't be able to unless you confirm that the code is licensed under terms that permit reuse.

This file also links to the licence, which 404s:
https://github.com/webcomponents/webcomponentsjs/blob/master/CONTRIBUTING.md

addyosmani:
Mostly self-explanatory. Verified with @tjsavage that this is the correct license information to include.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,27 @@
+// Copyright (c) 2015 The Polymer Authors. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


addyosmani:
Actually, let me just use markdown.

PR #205, opened 2015-02-17
fix(ShadowCSS): fix makeScopeMatcher() RegExp

vicb:
It seems like the right thing to do. However I don't have a test for this.


jmesserly:
yeah, it doesn't look like the old code can possibly be right :). lgtm

Issue #219, opened 2015-03-04
Polymer.import callback gets called even when path to HTML file is invalid


akutamruta:
Polymer.import(['/path to html file'], function callback () {
   console.log('hello from loaded import');
});

If the path mentioned in invalid, then there is an error loading the HTML import. In this case the callback should not be called. In the above example, "hello from loaded import" will get printed on the console when path is invalid

IntranetFactory:
I'm not sure if the change is ok - in the original version the callback gets the errorImports and can handle the error condition. With your change the callback is not called for errors and therefore won't be able to react on errors any longer.

akutamruta:
Makes sense. I agree that the callback should be responsible for handling the error condition. I will make the change in polymer.js. Thanks :)

addyosmani:
Thanks @akutamruta. Agree we should try to ensure the callback continues to get the `errorImports` included before landing this.

garlicnation:
This is working as intended. In your callback you can check the argument's errorImports and loadedImports arguments to determine if the import loaded successfully.

PR #220, opened 2015-03-04
Fixed issue #219 related to polymer import callback getting called

akutamruta:
Fix for #219 ( #220 ) : added the condition in checkDone() function to check if there is an error encountered in loading the import. If so, then I am preventing the callback() from getting called. The function call of callback()  in checkDone() is internally triggering the whenReady function in polymer.js which in turn calls the web components' callback(mentioned in Polymer.import) while flushing the ready queue.

### DIFF
--- a/src/HTMLImports/base.js
+++ b/src/HTMLImports/base.js
@@ -118,11 +118,14 @@ function watchImportsLoad(callback, doc) {
   var parsedCount = 0, importCount = imports.length, newImports = [], errorImports = [];
   function checkDone() {
     if (parsedCount == importCount && callback) {
-      callback({
-        allImports: imports,
-        loadedImports: newImports,
-        errorImports: errorImports
-      });
+	
+		if(errorImports.length == 0) {
+			callback({
+			allImports: imports,
+			loadedImports: newImports,
+			errorImports: errorImports
+		  });
+		}      
     }
   }
   function loadedImport(e) {


addyosmani:
Requires a little more work based on the comments in #219. Thanks for taking a stab at this.

akutamruta:
My pleasure @addyosmani . Was investigating the issue in polymer.js and found the place where it needs to be handled. I have made a fix but I think it might/ might not need enhancement based on the answer to the following question. 
Should Polymer continue loading imports if it encounters  an error while loading a particular import ? Say we have 3 imports and import 1 loads, import 2 has an error while loading. Should import 3's callback be called ? Or should the loading process stop after the error is encountered. Going logically, I think the callback for import 3 should be called. What's the requirement ?

addyosmani:
Good question. My understanding is the callback for the last import should still be called (if the second import errors out), however @jmesserly would have a clearer picture of the exact desired behaviour here.

garlicnation:
@akutamruta I don't have much visibility into this issue, but could the fix just be for Polymer(0.5, I'm assuming?) to check the argument for loaded/errorImports, and throw an error in that case? This change feels like it will drop information that would otherwise have been useful and could potentially allow the importer to recover from the error condition.

akutamruta:
Fixed the issue in polymer. Here is the pull request for the polymer repo.  https://github.com/Polymer/polymer/pull/1267

addyosmani:
@akutamruta given the Polymer PR can we close this issue here?

akutamruta:
Yes. Closing it.

PR #227, opened 2015-03-08
Fix Node.isEqualNode and document.createTreeWalker

eeid26:
1) Node.isEqualNode returns false always.
2) document.createTreeWalker always throws exception.

Issues:
https://github.com/webcomponents/webcomponentsjs/issues/228
https://github.com/webcomponents/webcomponentsjs/issues/229

### DIFF
--- a/src/ShadowDOM/wrappers/Document.js
+++ b/src/ShadowDOM/wrappers/Document.js
@@ -44,7 +44,7 @@
 
   defineWrapGetter(Document, 'documentElement');
 
-  // Conceptually both body and head can be in a shadow but suporting that seems
+  // Conceptually both body and head can be in a shadow but supporting that seems
   // overkill at this point.
   defineWrapGetter(Document, 'body');
   defineWrapGetter(Document, 'head');
@@ -118,6 +118,19 @@
           '[name=' + JSON.stringify(String(name)) + ']');
     }
   });
+
+  var originalCreateTreeWalker = document.createTreeWalker;
+  var TreeWalkerWrapper = scope.wrappers.TreeWalker;
+  Document.prototype.createTreeWalker = function(root, whatToShow, filter, expandEntityReferences ) {
+
+    var newFilter;
+    if (filter && filter.acceptNode && typeof filter.acceptNode === 'function'){
+      newFilter = { acceptNode:function(node) { return filter.acceptNode(wrap(node)); }  }
+    }
+
+    return new TreeWalkerWrapper(originalCreateTreeWalker.call(unwrap(this), unwrap(root),
+      whatToShow, newFilter, expandEntityReferences ));
+  };
 
   if (document.registerElement) {
     var originalRegisterElement = document.registerElement;
@@ -264,6 +277,7 @@
     'createEventNS',
     'createRange',
     'createTextNode',
+    'createTreeWalker',
     'elementFromPoint',
     'getElementById',
     'getElementsByName',
@@ -337,13 +351,6 @@
     'createHTMLDocument',
     'hasFeature',
   ]);
-
-  var originalCreateTreeWalker = document.createTreeWalker;
-  function createTreeWalker(root, whatToShow, filter, entityReferenceExpansion){
-    return originalCreateTreeWalker.call(document,unwrap(root), whatToShow, filter, entityReferenceExpansion);
-  }
-
-  document.createTreeWalker = createTreeWalker; // override
 
   scope.adoptNodeNoRemove = adoptNodeNoRemove;
   scope.wrappers.DOMImplementation = DOMImplementation;
--- a/src/ShadowDOM/wrappers/Document.js
+++ b/src/ShadowDOM/wrappers/Document.js
@@ -121,11 +121,21 @@
 
   var originalCreateTreeWalker = document.createTreeWalker;
   var TreeWalkerWrapper = scope.wrappers.TreeWalker;
-  Document.prototype.createTreeWalker = function(root, whatToShow, filter, expandEntityReferences ) {
+  Document.prototype.createTreeWalker = function(root,whatToShow,
+                                                 filter,expandEntityReferences ) {
 
-    var newFilter;
-    if (filter && filter.acceptNode && typeof filter.acceptNode === 'function'){
-      newFilter = { acceptNode:function(node) { return filter.acceptNode(wrap(node)); }  }
+    var newFilter = null; // IE does not like undefined.
+
+    // Support filter as a function or object with function defined as acceptNode.
+    // IE supports filter as a function only. Chrome and FF support both formats.
+    if (filter){
+      if (filter.acceptNode && typeof filter.acceptNode === 'function'){
+        newFilter = {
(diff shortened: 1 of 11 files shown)


addyosmani:
Can you include a description and link to the issue this addresses?

src/ShadowDOM/wrappers/Document.js
```diff
@@ -338,6 +338,13 @@
     'hasFeature',
   ]);
 
+  var originalCreateTreeWalker = document.createTreeWalker;
+  function createTreeWalker(root, whatToShow, filter, entityReferenceExpansion){
+    return originalCreateTreeWalker.call(document,unwrap(root), whatToShow, filter, entityReferenceExpansion);
+  }
+
+  document.createTreeWalker = createTreeWalker; // override
```
jmesserly: hey, this is not quite the right way to do this, but it's close :)

Document is really subtle, because `window.document` cannot be overridden, so we put methods on both the Document wrapper and then a forwarder on the global document instance.

Could you move your implementation into the `mixin(Document.prototype, { ... })` block above in this file, then add this method name under `forwardMethodsToWrapper`

And perhaps add a test case along these lines:
```
// should work
wrap(document).createTreeWalker( ... )
// this too
document.createTreeWalker(...)
```

Wrapped documents need to have the `createTreeWalker` method, because they can be returned in various other ways (like `ownerDocument`)

jmesserly: oops, also the implementation will look something like:
```
    createTreeWalker: function() {
      return originalCreateTreeWalker.call(unwrap(this), unwrap(root), whatToShow, filter, entityReferenceExpansion);
    },
```

(note the "unwrap(this)" for the document itself)

src/ShadowDOM/wrappers/Node.js
```diff
@@ -682,6 +683,10 @@
                                                   unwrapIfNeeded(otherNode));
     },
 
+    isEqualNode:function(otherNode){
```
jmesserly: style: this appears to be missing some spaces

jmesserly:
Added some comments, otherwise this looks really good, thanks for sending this.

src/ShadowDOM/wrappers/Document.js
```diff
@@ -338,6 +338,13 @@
     'hasFeature',
   ]);
 
+  var originalCreateTreeWalker = document.createTreeWalker;
+  function createTreeWalker(root, whatToShow, filter, entityReferenceExpansion){
+    return originalCreateTreeWalker.call(document,unwrap(root), whatToShow, filter, entityReferenceExpansion);
```
jmesserly: one more thought: do we need to handle wrapping on the other arguments or the result here? In other words, if those APIs take Nodes, they need to deal with wrapping/unwrapping too.

jmesserly: From https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker
it looks like the filter should be wrapped, something like:
```js
// Make sure we only pass wrapped nodes to the user's filter method
var newFilter = function(node) { return filter(wrap(node)); }
```

... and then the tree walker API's need to be wrapped too:
https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker

those methods should all return wrapped nodes. Something like this:

```js
// wrappers/TreeWalker.js
function TreeWalker(impl) {
  setWrapper(this, impl);
}
mixin(TreeWalker.prototype, {
  nextNode: function() { return wrapIfNeeded(unsafeUnwrap(this).nextNode()); },
  ...
});

registerWrapper(...)
```

here's an example of the pattern: https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/wrappers/Selection.js

src/ShadowDOM/wrappers/Document.js
eeid26: I think the new checkin is closer to what you described.

src/ShadowDOM/wrappers/Node.js
eeid26: I could not find that missing space. I might be blind.

jmesserly: ah, I meant:
```diff
-isEqualNode:function(otherNode){
+isEqualNode: function(otherNode) {
```
space after colon and between close paren and open curly

src/ShadowDOM/wrappers/Selection.js
```diff
@@ -56,6 +53,15 @@
     }
   };
 
+  // Not all browsers support Selection.extend. Some versions of IE do not support extend, code that checks
+  // if extend exists in the Selection would fail the test if we define extend on the wrapper and it does not exist in
+  // the browser Selection object.
+  if (window.getSelection().extend){
```
jmesserly: also do you mind adding space here before '{' ?

src/ShadowDOM/wrappers/Selection.js
```diff
@@ -56,6 +53,15 @@
     }
   };
 
+  // Not all browsers support Selection.extend. Some versions of IE do not support extend, code that checks
```
jmesserly: which versions of IE?

src/ShadowDOM/wrappers/Selection.js
jmesserly: also, I think you could do: `if (OriginalSelection.prototype.extend) {` and avoid calling getSelection

src/ShadowDOM/wrappers/Selection.js
```diff
@@ -56,6 +53,15 @@
...
+  // if extend exists in the Selection would fail the test if we define extend on the wrapper and it does not exist in
```
jmesserly: fyi, the JS style guide this code used (https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml) normally has comments wrapping at 80 columns.

(apologies for nitpicking style. I don't personally feel it's a big deal, but a lot of folks do, so as a reviewer I try and remember to check for it.)

src/ShadowDOM/wrappers/Node.js
eeid26: now I understand what space you are referring to. I fixed it.

jmesserly:
A few minor style comments, otherwise this LGTM! Thank you!

src/ShadowDOM/wrappers/Selection.js
eeid26: I know IE11 on windows does not have it. That is what I am testing with. It seems also some one already discovered this and checks if IE is running and not test extend. I changed the test.

src/ShadowDOM/wrappers/Selection.js
eeid26: let me check the styling guide and format the code accordingly.

eeid26:
I tried to follow the styles, but the 80 chars rule is a tough one. 
I just need to find a small monitor :)

jmesserly:
Haha, yeah, the 80 can be feel a bit cramped at times. Thanks again. Merging.

PR #226, opened 2015-03-07
Patch importNode to work around IE11 bug.

kevinpschaaf:

PR #232, opened 2015-03-09
Remove shadowDom flag from webcomponents-lite

garlicnation:
Need to rebase after #226 lands before merging.

### DIFF
--- a/webcomponents-lite.js
+++ b/webcomponents-lite.js
@@ -34,7 +34,7 @@
       }
     }
     // log flags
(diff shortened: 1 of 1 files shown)


garlicnation:
@sorvell No reason to have this flag for -lite at all, right?

addyosmani:
@sorvell would be way more knowledgable than I on this, but afaics the flag isn’t necessary for -lite, no.

sorvell:
That's correct. Let's remove it.

Issue #235, opened 2015-03-10
[HTMLImports] IE11 sometimes fails to successfully import documents


sorvell:
Yields the following error:

SCRIPT5022: HierarchyRequestError
importer.js, Line: 128, Column: 3

azakus:
Is there a particular import that you see that with? Got a hunch that this is an element being appended to `<html>` that isn't a `<body>` or `<head>`

sorvell:
We're hitting this, for example, when running the Polymer 0.8 tests using WCT. They will just randomly stop with this error. It's possible that this only occurs in the context of using WCT.

garlicnation:
The problem is in createHTMLDocument. It starts behaving really badly when called in fast enough succession. We might be able to fix this in the tests by adding some delay between running test suites, and I've reproduced the IE bug in https://github.com/garlicnation/crashie.

Chrome runs that html just fine, IE either crashes or ends up with a HierarchyRequestError.

sorvell:
is it enough to guarantee that IE does only one of these creates per task?
If so, we could probably adjust the imports polyfill to do this for IE.

azakus:
If we can't make createHTMLDocument stable enough, we can say that IE is only 100% supported when using vulcanize.

IntranetFactory:
I've tested the code from https://github.com/garlicnation/crashie with IE 11 on Win 8.1 and it neither crashed IE nor did it cause a HierarchyRequestError. But it's slow, and every ~4,000 items it stops for 3-5 seconds.

garlicnation:
The IE update from March 10, 2015 resolves this issue.

brianblakely:
I was experiencing this issue in 0.7.22.  Suddenly, it just stopped happening.

In `TemplateImpl.decorate` at this line: `template.content.appendChild(child);`

Also in `makeDocument` at this line: `var doc = document.implementation.createHTMLDocument(IMPORT_LINK_TYPE);`

bedeoverend:
FWIW I'm getting this issue as well. on v0.7.22 and IE11 on Windows8.1

humana-fragilitas:
[...]

PR #236, opened 2015-03-10
Ie fixes galore.

garlicnation:
Add a retry loop to paper over the bug in #235


src/CustomElements/boot.js
```diff
@@ -57,6 +78,7 @@ if (!window.wrap) {
   }
 }
 
+
```
azakus: micronit: extra newline?

garlicnation: that should go into #226.

azakus: Done. Disregard this.

PR #237, opened 2015-03-11
Add URL toString() method

threeday0905:
Add URL toString() method to return synonym for href. (according to spec)

Current build:

    var u = new URL('b', 'http://a');
    console.log( u.toString() ); // "[object Object]""

After patch:

    var u = new URL('b', 'http://a');
    console.log( u.toString() ); // "http://a/b"

// currently I cannot find URL related test on tests folder, should I write some tests for this patch?

### DIFF
--- a/src/URL/URL.js
+++ b/src/URL/URL.js
@@ -467,6 +467,9 @@
   }
 
   jURL.prototype = {
+    toString: function() {
+      return this.href;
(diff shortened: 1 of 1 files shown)


threeday0905:
I use the wrong email to commit, I will close this PR.

PR #238, opened 2015-03-11
Add URL toString() method to return synonym for href.

threeday0905:
Add URL toString() method to return synonym for URL href. (according to URL spec)

Current build:

    var u = new URL('b', 'http://a');
    console.log( u.toString() ); // "[object Object]""

After patch:

    var u = new URL('b', 'http://a');
    console.log( u.toString() ); // "http://a/b"

// currently I cannot find URL related test on tests folder, should I write tests for this patch?

### DIFF
--- a/src/URL/URL.js
+++ b/src/URL/URL.js
@@ -467,6 +467,9 @@
   }
 
   jURL.prototype = {
+    toString: function() {
+      return this.href;
+    },
     get href() {
       if (this._isInvalid)
         return this._url;


addyosmani:
The original source for the URL polyfill exists over in https://github.com/Polymer/URL, which the team decided to just statically include. That source includes tests.

Imo, changes like this should probably be proposed upstream, but it also opens up some questions:

* Should changes like this land here or upstream? (my vote is upstream)
* Should we just be pulling this dep in via Bower as a dependency?
* Are we/the team fine with occasionally updating the statically checked in version of the dep for changes like this?.

Deferring to @azakus for his input.

azakus:
Upstream annevk/url points to Polymer/URL as a "more likely to be maintained" version, so I say we can probably just land the change here and keep Polymer/URL sync'd up.

azakus:
ping @garlicnation

addyosmani:
As we're okay with landing the changes here and keeping the upstream repo sync'd up, change LGTM to land in master.

Issue #110, opened 2014-12-09
support for ShadowRoot/Document activeElement


tjsavage:
<a href="https://github.com/sigmundch"><img src="https://avatars.githubusercontent.com/u/2049220?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [sigmundch](https://github.com/sigmundch)**
_Monday Jul 28, 2014 at 23:27 GMT_
_Originally opened as https://github.com/Polymer/ShadowDOM/issues/478_

----

It seems we don't have a polyfill for [document.activeElement][1]. Not sure if that was intentional.

Here is a simple example:
```html
<!-- <script src="ShadowDOM.min.js"></script> -->
<div></div>
<script>
document.addEventListener('click', function() {
  console.log(document.activeElement);
});

var div = document.body.querySelector('div');
var root = div.createShadowRoot();
root.innerHTML = '<input id="a"></input>';
</script>
```

This program should print `<div>` when clicking on the input element `#a` (this works with the native implementation in Chrome 37). When using the polyfill, the code prints the input element instead.

This came up in the context of [a bug reported by a Dart user][2].

[1]: http://www.w3.org/TR/shadow-dom/#active-element
[2]: https://code.google.com/p/dart/issues/detail?id=20165

bicknellr:
Working on this in #457.

PR #243, opened 2015-03-16
Add an activeElement getter for #110

garlicnation:


addyosmani:
LGTM.

jmesserly:
Hmmm. This doesn't respect the ShadowDOM encapsulation right? 
https://github.com/webcomponents/webcomponentsjs/issues/110
http://w3c.github.io/webcomponents/spec/shadow/#active-element

(apologies if I'm missing some context)

garlicnation:
This is causing a build error, backing out for now.

@jmesserly, no, it doesn't, but this seemed like a decent solution to prevent the workarounds needed from #110. I'd much prefer to fix this in a spec-compliant way.

addyosmani:
Ah. Sorry for not catching that this broke SD encapsulation. I should have during review.

garlicnation:
@jmesserly Thanks for the spec link. It should be easy to fix this up.

PR #245, opened 2015-03-18
Add support for the >>> selector in ShadowCSS

vicb:
">>>" is the new name for "/deep/"
see http://dev.w3.org/csswg/css-scoping-1/#deep-combinator

### DIFF
--- a/tests/ShadowCSS/html/combinators-shadow.html
+++ b/tests/ShadowCSS/html/combinators-shadow.html
@@ -40,6 +40,10 @@
       x-bar /deep/ .zot {
         color: white;
       }
+
+      x-bar >>> .zot-2 {
+        color: white;
+      }
     </style>
     <div class="foo">orange</div>
     <div class="foo2">green</div>
@@ -58,6 +62,7 @@
 
   <template id="x-zot">
     <div class="zot">Zot</div>
+    <div class="zot-2">Zot</div>
     <div class="zot-inner">inner</div>
   </template>
 
@@ -84,11 +89,18 @@
       var xZot = foo.shadowRoot.querySelector('x-bar').shadowRoot
           .querySelector('x-zot');
       var zot = xZot.shadowRoot.querySelector('.zot');
+      var zot2 = xZot.shadowRoot.querySelector('.zot-2');
 
       chai.assert.equal(getComputedStyle(zot).backgroundColor, 'rgb(0, 128, 0)',
-        '^ styles are applied (backgroundColor)');
+        '::shadow styles are applied (backgroundColor)');
       chai.assert.equal(getComputedStyle(zot).color, 'rgb(255, 255, 255)',
-        '^^ styles are applied (color)');
+        '/deep/ styles are applied (color)');
+
+      // TODO(vicb): remove the if when >>> get supported by browsers
+      if (window.ShadowDOMPolyfill) {
+        chai.assert.equal(getComputedStyle(zot2).color, 'rgb(255, 255, 255)',
+          '>>> styles are applied (color)');
+      }
 
       // NOTE: native import styles not shimmed under SD polyfill
       var unsupported = window.ShadowDOMPolyfill && HTMLImports.useNative;
--- a/tests/ShadowDOM/js/Document.js
+++ b/tests/ShadowDOM/js/Document.js
@@ -216,6 +216,9 @@ htmlSuite('Document', function() {
 
     assert.equal(aa1, document.querySelector('div /deep/ aa'));
     assert.equal(bb, document.querySelector('div /deep/ bb'));
+
+    assert.equal(aa1, document.querySelector('div >>> aa'));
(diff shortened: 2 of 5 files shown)


addyosmani:
Please note that as of [this](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/hRw781MV3mE) intent to ship, Blink/Chrome has concluded that it doesn't seem like a good idea to proceed with `>>>`, given that we're not even sure we should keep shipping the `/deep/` combinator.

This suggests that we should take time before considering landing support for `>>>` in the polyfills. @dglazkov do you have any thoughts here?

dglazkov:
In abstract, I don't think it makes sense to couple Blink shipping decisions with those of the polyfill. The interesting question here is how are the shipping decisions for the polyfill project made? I would imagine that it should aspire to implement all of the currently-spec'd behavior and update as the specs are updated. If that's the case, if `>>>` is present in the spec, the polyfill should support it. How's my logical reasoning? :smiley_cat:

addyosmani:
Thanks for your input, @dglazkov!
 
> The interesting question here is how are the shipping decisions for the polyfill project made? I would imagine that it should aspire to implement all of the currently-spec'd behavior and update as the specs are updated. If that's the case, if >>> is present in the spec, the polyfill should support it.

In retrospect, if we intend to be a full-fidelity polyfill then we should aim to completely support the spec. Interested in what @garlicnation and @jmesserly think of this.

jmesserly:
:+1: my vote would be to support spec'd behavior + whatever backwards compat selectors are needed pragmatically.

also, changes here LGTM.

addyosmani:
After discussing this change at our last sync, we would love to be able to move ahead with it. @vicb Could you author tests to accompany this change? That's probably all that's still missing here.

addyosmani:
Friendly re-ping @vicb on the test coverage question above :)

vicb:
Thanks for the friendly reminder, I'll take a look tomorrow -(last time I've checked the tests I was a bit lost...)_

addyosmani:
If you have a question about any of the testing setup, please feel free to holler :) Appreciate you looking at this!

tests/ShadowCSS/html/combinators-shadow.html
```diff
@@ -84,11 +89,18 @@
...
+      var zot2 = xZot.shadowRoot.querySelector('.zot-2');
 
       chai.assert.equal(getComputedStyle(zot).backgroundColor, 'rgb(0, 128, 0)',
-        '^ styles are applied (backgroundColor)');
+        '::shadow styles are applied (backgroundColor)');
       chai.assert.equal(getComputedStyle(zot).color, 'rgb(255, 255, 255)',
-        '^^ styles are applied (color)');
+        '/deep/ styles are applied (color)');
+
+      // todo(vicb): remove the if when >>> get supported by browsers
```
vicb: is this right ?

tests/ShadowDOM/js/Document.js
```diff
@@ -230,14 +233,18 @@ htmlSuite('Document', function() {
 
     div.offsetHeight;
 
-    var list = document.querySelectorAll('div /deep/ aa');
```
vicb: is there a reason why some tests look duplicated in Document.js & Element.js (see below)

vicb:
@addyosmani I've pushed a tentative commit with some inline questions. Could you please provide some feedback when you get a chance to review it. Thanks.

tests/ShadowCSS/html/combinators-shadow.html
addyosmani: This looks fine to me.

tests/ShadowDOM/js/Document.js
vicb: @addyosmani should I remove duplicate tests before squashing the commits  in this PR ?

addyosmani: I'd like to wait for a quick confirm from @jmesserly on the dupe tests before we remove em :)

vicb: I'll open a separate issue so that this PR could me merged

vicb:
@addyosmani rebased & squashed, should be ready to go. Thanks for the review / feddback.

tests/ShadowDOM/js/Document.js
jmesserly: Hmm, I might be missing context. By "look duplicated" do you mean similar tests for Document.querySelectorAll vs Element.querySelectorAll?

`Document` has very unusual wrapping strategy (because the global `document` can't be patched to be the wrapped document node), so it's usually worth extra tests there.

vicb: @jmesserly Thanks for the feedback, I've just noticed it by looking more closely at the code. Sorry for the noise.

addyosmani:
Changes LGTM. Landing. Thanks for your work on the patch and tests, @vicb!

PR #250, opened 2015-03-19
Add WM and MO polyfills to lite for IE. Make build manifest match.

kevinpschaaf:


addyosmani:
Changes LGTM. Just one comment as we previously landed the URL polyfill in that position back in #197

garlicnation:
Unfortunately, we need the URL polyfill for #135 in IE.

addyosmani:
:+1:

PR #251, opened 2015-03-20
Fixes #168 - refer to correct file for-lite builds

### LINKED ISSUE #168: webcomponents-lite.js "release" version refers to itself as webcomponents.js
The latest versions of webcomponents lite in the "releases" directory that I've seen (at least v0.5.3 and v0.5.4) refer to file being loaded as webcomponents.js instead of webcomponents-lite.js. Here is line 15 from v0.5.4's webcomponent-lite.js:

  var file = "webcomponents.js";

It's also worth noting that this seems to have been fixed in webcomponents-lite.js in the "master" branch on line 17. However, the v0.5.4 and "master" versions of webcomponents-lite are quite different. The "master" version is streamlined to 78 lines and seems to define the modules to be included in webcomponents-lite.js. However, the v0.5.3 and v0.5.4 versions are ~1,750 lines and not modularized.

addyosmani:
This is a quick and dirty solution for #168. 

An alternative would be to use feature flags to determine whether someone is trying to build a custom elements + imports build (and just call the file -lite then). @garlicnation PTAL.

### DIFF
--- a/src/WebComponents/build-lite.json
+++ b/src/WebComponents/build-lite.json
@@ -1,5 +1,5 @@
 [
-  "build/boot.js",
+  "build/boot-lite.js",
   "../URL/URL.js",
   "../MutationObserver/MutationObserver.js",
   "../HTMLImports/build.json",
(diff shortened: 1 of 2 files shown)


addyosmani:
@garlicnation circling back on this now that we have the webcomponents-lite package. I'm sure there's a better solution possible than what I have here, but may be useful for us to think about https://github.com/webcomponents/webcomponentsjs/issues/168.

garlicnation:
This is fine for now. We need to think about how to split the repos up(likely using NPM) going forward, though.

addyosmani:
Sgtm!

PR #253, opened 2015-03-24
Add nuget package

adityapurwa:
Add a nuget nupkg that is uploaded to NuGet gallery so it can be installed to Visual Studio easily. See https://www.nuget.org/packages/WebComponentJS/


azakus:
@adityapurwa Can you describe what the contents of this file are? It seems rather huge. I'm guessing a compressed file of some kind?

adityapurwa:
This is the wrong file and the nupkg file is corrupt. I'll fix it first

PR #257, opened 2015-03-25
fix #256 -- DOM objects with numeric properties

### LINKED ISSUE #256: Error in Shadow Dom polyfill when name an integer
In the Shadow Dom polyfills, on the rare occasion that the `getGetter` and `getSetter` methods get a name parameter that is an integer, the following error is thrown:

`Error: Expected ';'`

Changing 
````
return hasEval && isIdentifierName(name) ? new Function("return this.__impl4cf1e782hg__." + name) : function() {
        return this.__impl4cf1e782hg__[name];
};
````

to 
````
return hasEval && isIdentifierName(name) ? new Function("return this.__impl4cf1e782hg__['" + name + "']") : function () {
        return this.__impl4cf1e782hg__[name];
};
````

and 
````
return hasEval && isIdentifierName(name) ? new Function("v", "this.__impl4cf1e782hg__." + name + " = v") : function(v) {
        this.__impl4cf1e782hg__[name] = v;
};
````

to 
````
return hasEval && isIdentifierName(name) ? new Function("v", "this.__impl4cf1e782hg__['" + name + "']" + " = v") : function (v) {
        this.__impl4cf1e782hg__[name] = v;
};
````

resolves this issue. I would assume the same issue would occur for the `getMethod`

jmesserly:
This fixes #256, an error caused by trying to wrap an integer property on a native DOM object. Apparently some embedded MS Office objects run on IE9 and have this behavior. The `\w` in the regexp was wrong because it matches digits. Also adds `$` because that is legal in JS identifiers.

Adds two tests: a unit test for isIdentifierName, and a test for wrapping an integer property. Both of them fail before the fix and pass after.

### DIFF
--- a/src/ShadowDOM/wrappers.js
+++ b/src/ShadowDOM/wrappers.js
@@ -150,7 +150,7 @@ window.ShadowDOMPolyfill = {};
   }
 
   function isIdentifierName(name) {
-    return /^\w[a-zA-Z_0-9]*$/.test(name);
+    return /^[a-zA-Z_$][a-zA-Z_$0-9]*$/.test(name);
   }
 
   // The name of the implementation property is intentionally hard to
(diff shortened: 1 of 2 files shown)


jmesserly:
R= @addyosmani @garlicnation mind taking a look?
Thanks to @chimon2000 for the bug report!

addyosmani:
Reviewed and change LGTM. I was going to ask for an OP verification that this resolves the original issue, but we've got that over in #256:

> I swapped out your sweet fix for mine and that looks to have resolved the issue (both in browser and in office) as well.

PR #259, opened 2015-03-26
build task for just MutationObserver

### LINKED ISSUE #258: Gulp build task for only MutationObserver
I'd like to use the MutationObserver polyfill, and don't really have a need for the rest of webcomponents.js. I'd like a build with just that.

jerryorr:
resolves #258


addyosmani:
@garlicnation in the past we've suggested folks are free to customise their own local checkouts to output the polyfills they need. 

Are we interested in updates to the build process for the just MO use case?

PR #264, opened 2015-04-02
fix issues with range and elements with shadow dom. 

eeid26:
issue: https://github.com/webcomponents/webcomponentsjs/issues/268

Before supporting contenteditable with elements that have shadow dom, the range selection needs to work with those kind of elements.

There are issues with the current Range implementation.
Some of these issues

1) If the host has a shadow dom like
```html
<div id="container">
    <content></content>
</div>
```
the setStart and setEnd will fail if the node is referencing the host node. The child nodes in the host might look like
```html
  <div>
       <span>one</span>
       <span>two</span>
       <span>three</span>
  </div>
```
But the host is really rendered as 
```html
  <div>
       <div id="container">
             <span>one</span>
             <span>two</span>
              <span>three</span>
        </div>
  </div>
```
So setting the range,setStart(host,2) or setEnd will fail with no child exist at offset 2.

2) The current implementation also returns shadow dom nodes in the commonAncestorContainer when the host is the start and end node. The spec suggests to never reveal nodes in shadow dom. That is what chrome native implementation does.

The contenteditable support for elements with shadow dom depends on the range working properly with elements with shadow dom.

The PR fixes some of these issues and builds on the Range unit tests to get us one step closer to support the editing api. ( document.execCommand and others)


jmesserly:
yeah, originally contentEditable was punted as being very hard to polyfill correctly.

That said it looks like you put a lot of work into this, including excellent tests. I'm happy to review. CC @garlicnation @addyosmani  since this is a decent sized new feature.

Any idea if the tests work cross browser?

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
 
   var OriginalRange = window.Range;
 
+  var ShadowRoot = scope.wrappers.ShadowRoot;
+
+  function isElement(node){
```
jmesserly: style note: it looks like spaces are missing between `){` ... I see this in a lot of places.

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+    return node && node.nodeType == Node.ELEMENT_NODE
+  }
+
+  function getHost(node){
+    while(node && node.parentNode){
+      if (node.parentNode && node.parentNode instanceof ShadowRoot){
+        return node.parentNode.host;
+      }
+      node=node.parentNode;
```
jmesserly: space also missing here `node = node.parentNode;`

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+    }
+  }
+
+  function hostNodeToShadowNode(refNode,offset){
+    if (refNode.shadowRoot){
+      // Note: if the refNode is an element, then selecting a range with and offset
+      // equal to refNode.childNodes.length+1 is valid. That is why calling Math.min
+      // is necessary to make sure we select valid children.
+      var child = refNode.childNodes[Math.min(refNode.childNodes.length-1, offset)];
```
jmesserly: space around `length-1` and `length>0` below

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+  }
+
   function Range(impl) {
     setWrapper(impl, this);
   }
   Range.prototype = {
     get startContainer() {
-      return wrap(unsafeUnwrap(this).startContainer);
+      // Never return a node in the shadow dom.
+      return shadowNodeToHostNode.call(this,"startContainer");
```
jmesserly: also space after `this,"` ... you get the idea :)

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
     },
     get endContainer() {
-      return wrap(unsafeUnwrap(this).endContainer);
+      return shadowNodeToHostNode.call(this,"endContainer");
     },
     get commonAncestorContainer() {
-      return wrap(unsafeUnwrap(this).commonAncestorContainer);
+      return shadowNodeToHostNode.call(this,"commonAncestorContainer");
```
jmesserly: instead of `.call(this)`, you could pass in `this` as an argument? i think it might be a little faster

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+          }
+        }
+      }
+    }
+    return refNode;
+  }
+
+  function shadowNodeToHostNode(rangeNodeFieldName){
+    var node = wrap(unsafeUnwrap(this)[rangeNodeFieldName]);
+    var host = getHost(node);
```
jmesserly: maybe: `return getHost(node) || node;`?

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+  }
+
+  function hostNodeToShadowNode(refNode,offset){
+    if (refNode.shadowRoot){
+      // Note: if the refNode is an element, then selecting a range with and offset
+      // equal to refNode.childNodes.length+1 is valid. That is why calling Math.min
+      // is necessary to make sure we select valid children.
+      var child = refNode.childNodes[Math.min(refNode.childNodes.length-1, offset)];
+      if (child){
+        var insertionPoint = ShadowDOMPolyfill.getDestinationInsertionPoints(child);
```
jmesserly: `scope` instead of `ShadowDOMPolyfill`

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
   var OriginalRange = window.Range;
 
+  var ShadowRoot = scope.wrappers.ShadowRoot;
+
+  function isElement(node){
+    return node && node.nodeType == Node.ELEMENT_NODE
+  }
+
+  function getHost(node){
+    while(node && node.parentNode){
```
jmesserly: fyi, I think this method can just be:
```js
var getTreeScope = scope.getTreeScope;

function getHost(node) {
  var root = getTreeScope(node).root;
  if (root instanceof ShadowRoot) return root.host;
  return null;
}
```

TreeScope caches a quick way to jump to the root of the current tree (either Document, DocumentFragment, or ShadowRoot)

src/ShadowDOM/wrappers/Range.js
```diff
@@ -20,23 +20,69 @@
...
+    if (refNode.shadowRoot){
+      // Note: if the refNode is an element, then selecting a range with and offset
+      // equal to refNode.childNodes.length+1 is valid. That is why calling Math.min
+      // is necessary to make sure we select valid children.
+      var child = refNode.childNodes[Math.min(refNode.childNodes.length-1, offset)];
+      if (child){
+        var insertionPoint = ShadowDOMPolyfill.getDestinationInsertionPoints(child);
+        if (insertionPoint.length>0){
+          var parentNode = insertionPoint[0].parentNode;
+          if (isElement(parentNode)){
```
jmesserly: could this be `parentNode.nodeType == Node.ELEMENT_NODE` or maybe `parentNode instanceof Element`? in other words, do you need the function?

src/ShadowDOM/wrappers/Range.js
jmesserly: actually on second thought, it might be faster to:
```js
get commonAncestorContainer() {
  return shadowToHostNode(unsafeUnwrap(this).commonAncestorContainer);
}
```
... the existing code tries to use `obj.prop` rather than `obj["prop"]` in some places, for better performance in some JS engines. Granted, it's probably not a big difference nowadays (and the existing code isn't super consistent on this).

tests/ShadowDOM/js/Range.js
```diff
@@ -11,76 +11,525 @@
...
+  var nativeCustomElementPrefix = "range-native-element-";
+  var customElementIndex=0;
+  var nativeCustomElementIndex=0;
 
-  teardown(function() {
-    if (div && div.parentNode)
-      div.parentNode.removeChild(div);
-    div = undefined;
+  setup(function() {
+    isNativeShadowDomSupported = !!unwrap(document.createElement('div')).createShadowRoot;
```
jmesserly: ah, clever. So you can compare native to polyfilled. That's neat.

one thing I wanted to do at some point, was make all of the SD polyfill tests work on native too. That way we can compare. I wouldn't necessarily assume native is right, if they differ, but would be nice to have a test suite that works on both.

anyway no action needed, just wanted to share some thougths

jmesserly:
I left some comments, mostly style. The core idea looks good to me. Awesome work!

src/ShadowDOM/wrappers/Range.js
eeid26: Do you have a tool to check the style? I am using Intellij and I am trying to set the style rules to detect webcomponents code style, but it seems I am still off.

jmesserly: found this post from our very own @addyosmani a few days ago:
https://medium.com/@addyosmani/auto-formatting-javascript-code-style-fe0f98a923b8

I haven't tried it on this code base but I'd be happy to accept auto reformatting with that tool (I can try too). Yeah, the code here might not be perfectly consistent, since webcomponentsjs is a few code bases merged together. Maybe we can reformat it all to be consistent. Or at least ShadowDOM codebase

eeid26: this utility is very good. The Range.js now has no style errors according to jscs. I also fixed the styles of some of the code I added before. SD has 274 google code style errors according to jscs. Maybe every time a file gets changed the developer who changes it could fix the styles.

src/ShadowDOM/wrappers/Range.js
eeid26: got it. I am still trying to understand what all the SD code is doing.

tests/ShadowDOM/js/Range.js
eeid26: Actually I already started working on the document.execCommand and I found my self needing some of the code in the Range tests, so I moved those functions to a ShadowDomTestUtils.js to allow me to create custom and standard elements with different shadow dom setup native and polyfill mode. I was also thinking other unit tests can use those functions to setup the tests and can test dynamically what the result of the native implementation. 

All the changes you suggested are checked in.

eeid26:
to answer you question 
"Any idea if the tests work cross browser?"
I tested the code and all webcomponents tests pass on windows 7 IE11, chrome and FF.

addyosmani:
The idea here looks really solid. I'll plan to review, x-browser test and check if this introduces any regressions at all early next week. Nice work @eeid26 and thanks so much for the detailed review @jmesserly!

eeid26:
I ran the tests on OSX 10.9/safari@7.0 in Sauce Labs
https://saucelabs.com/tests/9006568278a74701ac7c17a571b80064

one test failed 
OSX 10.9 safari 7.0      ✖ html/on-unload-test.html
 
It has nothing to do with the new code.

After this PR is merged, I will create another PR for the changes that will allow all the webcomponents tests to run with WCT (local and remote). The change will allow us to run the tests with WCT or without.

addyosmani:
> After this PR is merged, I will create another PR for the changes that will allow all the webcomponents tests to run with WCT (local and remote). The change will allow us to run the tests with WCT or without.

That would be amazing. Thank you!

jmesserly:
The latest diff LGTM :+1:

eeid26:
Are you waiting for me to make any other changes for the merge?

addyosmani:
We're waiting on a second review. We'll try to get this done shortly :)

src/ShadowDOM/wrappers/Range.js
addyosmani: I would be in favor of adding jscs linting/formatting to our contribution workflow. Thanks for updating range.js using it. Once this change lands, we can get a little more formal in our use of it.

addyosmani:
Source LGTM. Great work! At our last team meeting we discussed benchmarking changes to the polyfills against https://github.com/polymerlabs/benchmarks to avoid perf regressions. Once we've tested this against the PR, I'm :+1: on landing.

addyosmani:
Benchmark comparisons.

Before this change:

screen shot 2015-04-10 at 13 28 38

After:

screen shot 2015-04-10 at 13 37 40

screen shot 2015-04-10 at 13 28 48

My reading is that the change in numbers is sufficiently low to not raise concerns about perf regressions. 

Landing this change. Thanks so much for your contribution, @eeid26!

eeid26:
Thank you for the doing the merge. 

I am interested in the editing api (https://dvcs.w3.org/hg/editing/raw-file/tip/editing.html) support in polyfill shadow dom and eventually in Shady dom using polymer. 

After playing around with polymer 0.8 yesterday.
I am not sure about the future of the shadow dom polyfill in polymer future?
It looks to me that the Shady dom is replacing the Shadow dom polyfill. So now making the shadow dom polyfill work with elements that have polyfill shadow dom is just a practice implementation for implementing it in Shady dom or maybe create a separate add on for Shady dom.
Is this correct?

FYI, I was really disappointed to see this api in polymer 0.8
Polymer.dom(parent).appendChild(node)
Polymer.dom(parent).insertBefore(node, beforeNode)
Polymer.dom(parent).removeChild(node)
Polymer.dom(parent).querySelector(selector)
Polymer.dom(parent).querySelectorAll(selector)
Polymer.dom(parent).childNodes
Polymer.dom(node).parentNode
Polymer.dom(contentElement).getDistributedNodes()
Polymer.dom(node).getDestinationInsertionPoints()
Polymer.dom.flush()

Polymer magic was that it was not introducing any new api to manipulate the dom, its the same api developers are used to (document.appendChild, document.childNodes..etc) and now its not. I understand the goal was to improve performance, but in my opinion giving up the api was a very high cost. Insertion points and node distribution was all hidden from anyone who does not want to know about them, It was something that just worked.

nazar-pc:
@eeid26, I'm fighting with Range for few hours already (not even the first time approaching it) and now I believe this is actually a bug in polyfill, which is tightly related to this PR.

My environment is pretty complex (TinyMCE with full Shadow DOM support), but looks like following is a tests case that reproduces an issue: https://github.com/webcomponents/webcomponentsjs/compare/master...nazar-pc:range-test-fails

Would you be so kind to take a look at it? It seems that in test case added `range.startContainer` and `range.endContainer` are not returned correctly after setting with `range.setStart()` and `range.setEnd()`, which is exactly what I observe with TinyMCE. It happens after `getHost()`, since internally `startContainer` and `endContainer` are correct.

If someone else can take a look, please, do so)

eeid26:
I will look into this.
But I remember there was lots of issues with the range and the polyfill

eeid26:
I looked at your test changes and why it's failing. 
With your test changes, the test "div - <content> wrapped nested in Shadow DOM" fails in testRangeWith3SpansHTML.

Maybe the following code and comments can explain why your changes to the tests do not pass.


```javascript
function testRangeWith3SpansHTML(host) {
   .
   .
   .
    // create span0TextNode and span2TextNode for test readability.
    // Test selecting the text nodes inside the spans
    var span0TextNode = span0.childNodes[0];
    var span2TextNode = span2.childNodes[0];
    range.setStart(span0TextNode, 1);
    range.setEnd(span2TextNode, 1);
    selection.removeAllRanges();
    selection.addRange(range);
    assert.strictEqual(range.toString(), "neTwoT");

    // Unwrap the range.
    var unwrapedRange = ShadowDOMPolyfill.unsafeUnwrap(range);
    if (unwrapedRange) {
      // This is ok with the unwrapped range.
      assert.strictEqual(unwrapedRange.startContainer, ShadowDOMPolyfill.unwrapIfNeeded(span0TextNode));
      assert.strictEqual(unwrapedRange.endContainer, ShadowDOMPolyfill.unwrapIfNeeded(span2TextNode));
    }

    // @nazar-pc The change you made fails on the next code line.
    // If the text node is in the shadow dom, the polyfill returns the host,
    // see the comment in Range.js
    // Never return a node in the shadow dom.
    // The wrapped and unwrapped ranges will return the same toString value, but
    // they do not have the same startContainer, endContainer nor the offsets.
    // There are no specs really on how to deal with the range and shadow dom. But returning
    // nodes in the shadow dom and start editing them may not be the expected behaviour.
    assert.strictEqual(range.startContainer, span0TextNode);
    assert.strictEqual(range.endContainer, span2TextNode);

}
```

nazar-pc:
This is roughly what I thought about it. In this kind of scenarios I'm trying to make polyfill matching existing behavior of Chromium implementation, which is logical.
And Chromium indeed returns node from Shadow DOM, which makes more sense and is by far more useful actually.

Shouldn't `range.startContainer` always return what was set in `range.setStart()`? I need to find a way to fix this issue. Simple enough test case was the first step, solution for it is the next.

eeid26:
I really tried to match what chrome was doing in the native implementation at that time since there are no specs.  It seems now it's different.

nazar-pc:
Haven't fixed TinyMCE issue yet, but mentioned test case with fix for it in #589, PTAL

nazar-pc:
Indeed, it fixes the issue with content insertion in TinyMCE)

PR #270, opened 2015-04-06
Added possibility for recursive calls via callback for forDocumentTree function

edrachev:
When components has big import tree, it randomly falls at line:
<code>if (processingDocuments.indexOf(doc) >= 0) {</code>
with error "processingDocuments is null"
The reason is a recursive call of function forDocumentTree  via callback
Changed forDocumentTree function to work with such callback


jmesserly:
LGTM. Thank you!

PR #303, opened 2015-05-13
Namespace flags

kevinpschaaf:
Namespace query-string flags to (attempt) to avoid conflict with user query strings or other libraries.


src/CustomElements/CustomElements.js
```diff
@@ -18,7 +18,10 @@ var flags = {};
 if (!flags.noOpts) {
   location.search.slice(1).split('&').forEach(function(o) {
     o = o.split('=');
-    o[0] && (flags[o[0]] = o[1] || true);
+    var m;
```
garlicnation: Can you de-obfuscate these variable names?

garlicnation:
LGTM.

PR #307, opened 2015-05-20
Prevent register() from throwing exceptions when nativeConstructor is undefined

0x24a537r9:
In certain instances, register may be called with an undefined nativeConstructor thus throwing an exception when .prototype is not defined. If the native object does not exist, wrapping it makes no sense so we should just proceed rather than break all subsequent element registration.

### DIFF
--- a/src/ShadowDOM/wrappers.js
+++ b/src/ShadowDOM/wrappers.js
@@ -256,6 +256,9 @@ window.ShadowDOMPolyfill = {};
    *     properties from an instance object.
    */
   function register(nativeConstructor, wrapperConstructor, opt_instance) {
+    if (nativeConstructor == null) {
+      return;
+    }
     var nativePrototype = nativeConstructor.prototype;
     registerInternal(nativePrototype, wrapperConstructor, opt_instance);
     mixinStatics(wrapperConstructor, nativeConstructor);


jmesserly:
ideally, if something doesn't exist, we'd have caught it before this point. For example, [HTMLAudioElement](https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/wrappers/HTMLAudioElement.js#L21)
```js
  var OriginalHTMLAudioElement = window.HTMLAudioElement;

  if (!OriginalHTMLAudioElement) return;
```
Usually the ShadowDOM/wrappers/... files will do stuff and then call registerWrapper, the other stuff they do first might have side effects, so the registerWrapper call could be too late to bail. I think adding checks in the relevant element(s) is more consistent with the existing code. Do you know what type was missing?

jmesserly:
On second thought, I suppose this fix is better than blowing up :)

Looked through the wrappers and most of them this would be a safe point to bail. A few would still blow up like Selection (https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/wrappers/Selection.js#L61) or Window, but this fix wouldn't make that any worse. If it's fixing something for you then LGTM :+1:

PR #310, opened 2015-05-22
Ensure newly created templates have content

kevinpschaaf:
Part of fix for Polymer/polymer#1500

### DIFF
--- a/src/Template/Template.js
+++ b/src/Template/Template.js
@@ -27,10 +27,10 @@ if (typeof HTMLTemplateElement === 'undefined') {
     HTMLTemplateElement.decorate = function(template) {
       if (!template.content) {
         template.content = template.ownerDocument.createDocumentFragment();
-        var child;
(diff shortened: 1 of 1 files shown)


azakus:
My only concern is that this will break the `decorate` function for templates that were created by script before `DOMContentLoaded`.

@garlicnation thoughts?

kevinpschaaf:
@azakus the `content` created by decorate at DOMContentLoaded won't stomp on `content` created by this `createElement` addition (it existence checks here: https://github.com/webcomponents/webcomponentsjs/blob/ensure-tmpl-content/src/Template/Template.js#L28).

kevinpschaaf:
I'll make a change so that template children are moved into `content` even when the template is created imperatively, so that you could `innerHTML = ...` an imperatively-created template and have decorate move them into the content at DOMContentLoaded time.

garlicnation:
Merged.

PR #314, opened 2015-05-23
Make sure createElement normalizes input when checking custom element defintions

### LINKED ISSUE #313: [CustomElements]: custom elements fail to upgrade when registered with uppercase in name and extends is used
An element registered as "my-Element" (note the uppercase) and extending 'button' (or any native element)  will not upgrade when `document.createElement('button', 'my-Element')` is used; instead the user must lowercase the name.

See: http://jsbin.com/kulaqamube/1/edit?html,console,output

azakus:
Fixes #313 [CustomElements]: custom elements fail to upgrade when registered
with uppercase in name and extends is used


azakus:
TBR

PR #318, opened 2015-05-26
Fixes #317 - Ensure super class methods are added to GeneratedWrapper base classes as methods instead of a value with get/set

### LINKED ISSUE #317: Cannot call getBBox() on wrapped SVGTextElement in Internet Explorer 11: Invalid calling object
webcomponents.js v0.7.1
Creating a more specific ticket of issue #186 as it's not related to any charting libraries.
Created a tiny code example at http://embed.plnkr.co/XvdXdq/preview:
    
      var svgNode = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
      var textNode = document.createElementNS('http://www.w3.org/2000/svg', 'text');
      textNode.textContent = 'Hello world';
      textNode.setAttribute('y', 12);
      svgNode.appendChild(textNode);
      document.body.appendChild(svgNode);
      console.log(textNode.getBBox());

textNode.getBBox() throws a "Invalid calling object" in Internet Explorer 11, possibly due to cached pointer issue described in [Microsoft documentation](https://msdn.microsoft.com/library/gg622930(v=vs.85).aspx).  This essentially blocks much of SVG rendering with web components in IE.

Screenshot:
screen shot 2015-05-23 at 3 31 34 pm

talee:
Fixes issue #317 where IE couldn't call getBBox on SVGTextElement nodes.  The issue can be described as follows:

C extends B extends A, where A is the base class and C has new methods not in B or A.  We create a node C.  During registerInstanceProperties onto wrapped B, C's methods get added onto wrapped B as get/set.  Later in addForwardingProperties onto wrapped B, C's methods aren't added onto wrapped B because they've already been added incorrectly as get/set.

This occurs in IE because SVGTextElement.prototype defines getBBox and is separated from the "base" GeneratedWrapper (SVGElement) by some classes (SVGTextPositioningElement, SVGTextContentElement).  I think this issue would apply to other nodes that have similar inheritance/new method conditions described above (C, B, A).

The tests seem to be passing with the same numbers in Windows 7 IE 11, OS X Chrome/Firefox; I'm not sure what form of unit test would be good for this. Perhaps create 2 custom elements to recreate the C, B, A scenario.

### DIFF
--- a/tests/ShadowDOM/js/wrappers.js
+++ b/tests/ShadowDOM/js/wrappers.js
@@ -86,35 +86,67 @@ suite('Wrapper creation', function() {
     'video': 'HTMLVideoElement',
   };
 
+  var getCustomClasses = (function() {
+    function WrapperClass() {}
+    ShadowDOMPolyfill.registerWrapper(WrapperClass, function(){});
+    function BaseClass(){}
+    BaseClass.prototype = Object.create(WrapperClass.prototype, {
+      baseMethod: {
+        value: function(arg){return arg;}
+      }
+    });
+    function SuperClass(){};
+    // Set a method on a prototype whose parent prototype is not yet wrapped
+    SuperClass.prototype = Object.create(BaseClass.prototype, {
+      superMethod: {
+        value: function(arg){return arg;}
+      }
+    });
+    var Classes = {
+      WrapperClass: WrapperClass,
+      BaseClass: BaseClass,
+      SuperClass: SuperClass
+    };
+    return function(){
+      return Classes;
+    }
+  })();
+
+  test('Element prototype is first prototype of wrapped instance', function() {
+    // Before #317, prototype had to be retrieved with two calls:
+    // Object.getPrototypeOf(Object.getPrototypeOf(textNode))
+    var Classes = getCustomClasses();
+    var node = Object.create(Classes.SuperClass.prototype);
+    var wrappedNode = ShadowDOMPolyfill.wrap(node);
+    var wrappedPrototype = Object.getPrototypeOf(wrappedNode);
+    var customFnDescriptor = Object.getOwnPropertyDescriptor(
+        wrappedPrototype, 'superMethod');
+    assert.isDefined(customFnDescriptor,
+        'Prototype method should be on first prototype of instance');
+  });
+
+  // Super class is defined as a class above a base class, which is above an
+  // existing wrapped class (GeneratedWrapper).
+  test('Super class methods are wrapped as methods on base class', function() {
+    var Classes = getCustomClasses();
+    var node = Object.create(Classes.SuperClass.prototype);
+    var wrappedNode = ShadowDOMPolyfill.wrap(node);
+    ShadowDOMPolyfill.setWrapper(node, wrappedNode);
+    var wrappedPrototype = Object.getPrototypeOf(wrappedNode);
+    var customFnDescriptor = Object.getOwnPropertyDescriptor(
+        wrappedPrototype, 'superMethod');
+    assert.isFunction(customFnDescriptor.value,
+        'Super method should be a function');
+    assert.isUndefined(customFnDescriptor.get,
+        'Super method should not be get/set');
+    assert.isUndefined(customFnDescriptor.set);
+  });
 
   test('Br element wrapper', function() {
     var br = document.createElement('br');
     assert.isTrue('clear' in br);
     assert.isFalse(br.hasOwnProperty('clear'));
     assert.isTrue(Object.getPrototypeOf(br).hasOwnProperty('clear'));
-  });
-
-  // Super class is defined as a class above the existing wrapped class
-  // (GeneratedWrapper).
-  test('Super class methods are wrapped as methods on base class', function() {
-    // Set a method on a prototype whose parent prototype is not yet wrapped
-    SVGTextElement.prototype.customFn = function(arg) { return arg; };
-    var textNode = document.createElementNS(
-        'http://www.w3.org/2000/svg', 'text');
-    var wrappedPrototype = Object.getPrototypeOf(textNode);
-    var customFnDescriptor = Object.getOwnPropertyDescriptor(
-        wrappedPrototype, 'customFn');
-    // TODO: This assertion might qualify as a separate test? Not sure if
-    // accessing prototype via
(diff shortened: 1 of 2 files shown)


talee:
@azakus for review.

azakus:
@garlicnation @jmesserly

talee:
@garlicnation @jmesserly. I've added unit tests for these in wrappers.js.  They are passing in IE 11, OSX Chrome/Firefox/Safari 8.

jmesserly:
awesome, thank you @talee. I'm taking a look at this now, but those tests are really nice!

one very trivial issue, the spacing indent seems a bit off. Maybe like this?
```js
      if (typeof descriptor.value === 'function') { // NOTE: space between if and (
        if (allowMethod) { // NOTE: +2 indent to match other code in the file
          target[name] = getMethod(name);
        }
        continue;
      }
```

Regarding your TODO question in the test, yeah it's fine to assume wrapper prototype structure matches the real structure. Definitely would be alright to test `Object.getPrototypeOf(wrappedPrototype)`

Other than that I think it's a good fix, I'm going to play with it a little bit locally just to job my memory about the allowMethod flag.

tests/ShadowDOM/js/wrappers.js
```diff
@@ -94,6 +94,23 @@ suite('Wrapper creation', function() {
     assert.isTrue(Object.getPrototypeOf(br).hasOwnProperty('clear'));
   });
 
+  // Super class is defined as a class above the existing wrapped class (GeneratedWrapper).
+  test('Super class methods are wrapped as methods on base class', function() {
+      // Set a method on a prototype whose parent prototype is not yet wrapped
```
jmesserly: also noticed this is +4 indent instead of +2 like the other tests
(fwiw, I'm personally more a fan of +4, but existing style of the code was +2 ... I guess due to Google JS style guide)

jmesserly:
yeah, wow, great job tracking this down. To be honest, I don't fully understand why existing ShadowDOM impl is even doing what it's doing with registerInstanceProperties and opt_instance. It seems like a bug that registerInstanceProperties is ever used except with the real node instance (instead, opt_instance for registerInternal can be the prototype, or the prototype's prototype, etc...). Right now it seems like when building wrappers, everything on SVGTextElementPrototype will get copied up to its parent, SVGTextPositioningElementPrototype (the parent class), is that what you were seeing before your fix? Seems like this bug should be triggering all over the place. Maybe because we register most base class wrappers explicitly, we get lucky? I feel like this should be breaking more than just IE too, e.g. what I'm seeing stepping through your test in Firefox, I think it would've been broken as well.

Since the fix scales back the registerInstanceProperties craziness, it seems like a step in the right direction.

I do wonder a bit about a slightly more aggressive fix, refactoring how getWrapperConstructor works, so we don't end up thinking a prototype is an instance.

jmesserly:
Here's my alternate fix. Thoughts? Does it work for your case in IE too? This prevents ever trying to use the opt_instance except with the actual node object. https://github.com/webcomponents/webcomponentsjs/compare/master...jmesserly:fix_317

Stole your test :). Also fixed to use 80 col (not my preference, but I think is the style guide). It seems to pass Safari/Firefox/Chrome, I didn't try IE yet.

jmesserly:
here's a quick paste of the diff too:
```diff
diff --git a/src/ShadowDOM/wrappers.js b/src/ShadowDOM/wrappers.js
index f8099f5..0c77df6 100644
--- a/src/ShadowDOM/wrappers.js
+++ b/src/ShadowDOM/wrappers.js
@@ -101,7 +101,7 @@ window.ShadowDOMPolyfill = {};
   // Make sure they are create before we start modifying things.
   getOwnPropertyNames(window);
 
-  function getWrapperConstructor(node) {
+  function getWrapperConstructor(node, opt_instance) {
     var nativePrototype = node.__proto__ || Object.getPrototypeOf(node);
     if (isFirefox) {
       // HTMLEmbedElements will sometimes be [NS Object wrapper class]
@@ -121,7 +121,7 @@ window.ShadowDOMPolyfill = {};
     var parentWrapperConstructor = getWrapperConstructor(nativePrototype);
 
     var GeneratedWrapper = createWrapperConstructor(parentWrapperConstructor);
-    registerInternal(nativePrototype, GeneratedWrapper, node);
+    registerInternal(nativePrototype, GeneratedWrapper, opt_instance);
 
     return GeneratedWrapper;
   }
@@ -332,8 +332,13 @@ window.ShadowDOMPolyfill = {};
       return null;
 
     assert(isNative(impl));
-    return impl.__wrapper8e3dd93a60__ ||
-        (impl.__wrapper8e3dd93a60__ = new (getWrapperConstructor(impl))(impl));
+    var wrapper = impl.__wrapper8e3dd93a60__;
+    if (wrapper != null) {
+      return wrapper;
+    }
+
+    return impl.__wrapper8e3dd93a60__ =
+        new (getWrapperConstructor(impl, impl))(impl);
   }
 
   /**
```

talee:
@jmesserly Thanks for taking a detailed look.  Yes, you're right, most elements we get lucky because 1) the element has no methods of its own, 2) has a wrapped base class (doesn't have C extends B extends WrappedA requirement).  Let me fix the style (didn't notice the indent, thanks).

talee:
Creating SVGTextElement nodes felt specific, so I've changed the test to a generic super class and moved out the wrapped prototype assertion to its own test in 5af32cd1ad2772c5e35cbb981a9077446ea44d61.

talee:
Got WCT running on SauceLabs and with the last commit we have the following test coverage on JS files in tests/ShadowDOM:

Windows 8.1 chrome beta (618/2/0)       Windows 8 IE 10.0 (612/2/0)             Windows 8.1 IE 11.0 (612/2/0)         
Windows 8.1 firefox (615/2/0)           OS X 10.9 safari 7 (618/2/0)

As a side note: Not sure if there is a branch/config with WCT working, but I had to make a bunch of URL related changes to get things loading on SauceLabs.

talee:
Tests pass with the opt_instance == actual node object fix in conjunction with the method check inside installProperty.  Checked it in.  I think with both checks we've covered our bases at least for ShadowDOM tests.

jmesserly:
really like the new test! lgtm

PR #322, opened 2015-05-29
fixes #316, attribute deps were not recorded on the correct ShadowRenderer

### LINKED ISSUE #316: [ShadowDOM]: Filtered insertion points inside nested roots fail to distribute based on dom changes
Here's a repro: http://jsbin.com/naboya/1/edit.

Clicking the `Test` button toggles an attribute on the `#test` node and should cause it to toggle between the insertion points contained in its parent's shadowRoot. This fails.

Clicking the `Test2` button removes/adds the `#test` node. This appears to function correctly.

There's a clue in the fact that an incorrect distribution caused by clicking `Test` can be corrected by clicking `Test2` 2x to remove and re-add the node.

After some debugging, it appears that the renderer found for the `x-host` element does not properly report that it should invalidate based on the `foo` and `bar` attributes changing on children. Strangely, the `x-outer` element renderer _does_ report that it should invalidate on `foo` and `bar` which is also incorrect .

jmesserly:
Pretty straightforward, make sure we get the right renderer before calling poolDistribution. There are two places this can happen, so we check in both.

### DIFF
--- a/tests/ShadowDOM/js/test.js
+++ b/tests/ShadowDOM/js/test.js
@@ -17,6 +17,10 @@ suite('Shadow DOM', function() {
     el.offsetWidth;
     return unwrap(el).innerHTML;
   }
+  function getVisualText(el) {
+    el.offsetWidth;
+    return unwrap(el).textContent;
+  }
 
   function normalizeInnerHtml(s) {
     // IE9 - Even though the attribute name is stored as "checked" innerHTML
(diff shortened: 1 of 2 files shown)


jmesserly:
@sorvell @garlicnation mind taking a look?

tests/ShadowDOM/js/test.js
```diff
@@ -520,4 +524,40 @@ suite('Shadow DOM', function() {
...
+    div.innerHTML =
+        '<x-outer></x-outer>' +
+        '<template id="x-host">' +
+          'Foo: [<content select="[foo]"></content>] ' +
+          'Bar: [<content select="[bar]"></content>]' +
+        '</template>' +
+        '<template id="x-outer">' +
+          '<x-host id="host">' +
+            '<child foo>~Foo~</child> ' +
+            '<child foo bar>~Foo+Bar~</child> ' +
```
jmesserly: if the trailing space is removed here, things go really haywire, even reverting the ShadowRenderer fix. SD polyfill might be expecting text nodes between distributed elements (and they almost always are there in practice). Need to investigate more, but probably best to keep separate from this fix.

addyosmani:
Sorry for the delay in getting this reviewed @jmesserly. The changes here for ensuring we get the right renderer LGTM. The trailing space issue is imo, minor and could be addressed in a follow-up otherwise. 

cc @sorvell and @garlicnation in case they get another chance to look over this.

jmesserly:
Thanks Addy! Yeah I figured everyone was recovering from Polymer 1.0 launch fun :)

tests/ShadowDOM/js/test.js
garlicnation: Sounds like a good comment for this test.

garlicnation:
LGTM+nit.

addyosmani:
@jmesserly Do you want to go ahead and add in the comment from https://github.com/webcomponents/webcomponentsjs/pull/322#discussion_r31291288 and we can get this merged? Alternatively, happy to do this as part of a follow-up commit if that's more convenient.

tests/ShadowDOM/js/test.js
jmesserly: Sure. But to be clear, this comment is not related to this issue. The white space problem is a different bug: I reproduced it without the fix. I added the comment here on the pull request, to tie back to my comment in investigation notes:
https://github.com/webcomponents/webcomponentsjs/issues/316#issuecomment-106635479

But anyway, happy to add a comment to the code if it means landing this :)

jmesserly:
thanks for the reminder Addy, I didn't see the update. Yeah, I can add a comment, let me file a bug regarding the Text node distribution issue, so I can link it from the comment.

addyosmani:
:+1:

jmesserly:
Thanks for the nudge @garlicnation  ... it got me to actually file https://github.com/webcomponents/webcomponentsjs/issues/337 :-)

change updated. I'll rerun tests locally since it merged with lots of newer stuff, but otherwise sounds like it's good to merge

jmesserly:
tests look good

tests/ShadowDOM/js/test.js
jmesserly: added:
```js
// Trailing whitespace in this HTML is to match the original repro, and to
// work around: https://github.com/webcomponents/webcomponentsjs/issues/337
```

PR #328, opened 2015-06-05
Remove `Edge` from sniff in src/HTMLImports/base.js.

jdalton:
I've noticed [some sites](http://www.google.com/design/icons/) using webcomponents.js are failing to load in MS Edge. I've found switching the UA string off MS Edge to something else makes it work. It looks like it's related to the sniff in src/HTMLImports/base.js.

### DIFF
--- a/src/HTMLImports/base.js
+++ b/src/HTMLImports/base.js
@@ -70,7 +70,7 @@ Object.defineProperty(rootDocument, '_currentScript', currentScriptDescriptor);
   the polyfill and native implementation.
  */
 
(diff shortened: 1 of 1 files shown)


jmesserly:
nice, lgtm. I wonder if we can remove this one too? https://github.com/webcomponents/webcomponentsjs/blob/6abbfa78e5f56a84567c485c8e73e908b107250d/src/ShadowDOM/wrappers/Node.js#L352

jdalton:
I'm not sure. I figure we can try this one first then we can dig into the others later.

jdalton:
Thanks! Is there an eta for the next bump?

PR #333, opened 2015-06-14
Remove `<link>` tag from install polymer warning

thedeeno:
The polymer warning appears to be causing a GET `components/polymer/polymer.html` request (despite polymer already being on the page). For starters, the request itself is surprising. Worse, I don't have a `components` directory in my application so this request will always fail.

I'm not sure what's causing the error text to be evaluated this way, but removing it stops my application from making the mistaken request.

### DIFF
--- a/src/WebComponents/dom.js
+++ b/src/WebComponents/dom.js
@@ -80,8 +80,7 @@
     if (window.Polymer === polymerStub) {
       window.Polymer = function() {
         throw new Error('You tried to use polymer without loading it first. To ' +
-          'load polymer, <link rel="import" href="' +
-          'components/polymer/polymer.html">');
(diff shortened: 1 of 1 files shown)


azakus:
I can't repro this error making a GET request in isolation. It is just a string passed to a new Error object.

thedeeno:
Agreed. It's a surprising error. I haven't dug any deeper to isolate the processing step that's causing this to break. Likely a plugin or a library in my build pipeline. One thing I can say for certain though: changing the string resolves my issue.

I figured we could just reword this so it can't be misinterpreted by pre-processors. Too much?

TimvdLippe:
Closing this, as v2 no longer contains HTML imports. Sorry for the long wait 😢

PR #339, opened 2015-07-01
Define CustomEvents shim wherever it's missing, not just in IE

jre-g:
8d6a38a added a check confirming that we're running in IE before defining the `CustomEvent` shim. The purpose was to compensate for the fact that the previous `typeof window.CustomEvent !== 'function'` check behaved incorrectly in Safari, because their CustomEvent constructor had type `object`. However, this shim wasn't just required for Internet Explorer. It was also required for older versions of the Android Open Source Project browser, so the change broke support for that browser.

This PR instead just checks that `window.CustomEvent` is falsey, fixing the behaviour in the AOSP browser without breaking Safari.


jre-g:
Is there anything else that needs to be done for this PR? Anything I could do to help it along?

We had to include another copy of this definition when updating our polyfills, to work around the regression. It's not a big deal, but it would be cleaner if we could drop that the next time we update.

Thanks.

jmesserly:
This LGTM. Thanks for the ping!

azakus:
Ugh, this breaks IE 11. The check is insufficient because `window.CustomEvent` is indeed defined, just not usefully.

jmesserly:
Doh! :-\

Did the old check work? `isIE && typeof window.CustomEvent !== 'function'`?

Maybe we could do `<old check> || !window.CustomElement`, that would work in IE11 and AOSP browser?

jmesserly:
in the meantime maybe revert?

azakus:
Yeah I was testing that and it seems to work.

PR #340, opened 2015-07-02
Ensure attached & detached occur for moves in the same turn. Fixes #311

### LINKED ISSUE #311: removeChild followed by appendChild results in no detached/attached callbacks
Repro in test case here:  https://github.com/Polymer/polymer/pull/1591/files#r30874592

Removing `takeRecords` in following code results in no callbacks being called:

```
        domBind.parentElement.removeChild(domBind);
        // TODO(kschaaf): detached/attached not called if takeRecords isn't
        // called between remove & insert on polyfill... seems bad?
        CustomElements.takeRecords();
        container.appendChild(domBind);
```

kevinpschaaf:


sorvell:
LGTM

PR #341, opened 2015-07-03
Suppress runtime error if you load twice webcomponents-lite.js

zeitiger:
If you load webcomponents-lite.js twice you get an TypeError with 'flags.log.split is not a function'


azakus:
You probably have a bigger problem if you are loading webcomponentsjs twice.

addyosmani:
@zeitiger What use-cases are you running into where the polyfills are getting loaded in multiple times?

TimvdLippe:
Closing this PR due to age.

Issue #342, opened 2015-07-06
parseScript throws TypeError: null is not an object (evaluating 'script.parentNode.removeChild')


JanMiksovsky:
We have a complex single-page app that dynamically downloads page content and adds it to an existing document. Page content can include HTML Imports. Such imports used to work in Safari/Firefox, but at some point in the recent past, when the app tries to navigate to a new page, we're seeing the following exception:

```
TypeError: null is not an object (evaluating 'script.parentNode.removeChild')
```

The implicated call is in parseScript in parser.js. This routine appears to load a script and wait for the script to load. The script is associated with one of the HTML Imports that's being dynamically added to the document. We don't see this bug in Chrome with native HTML Imports, so this appears to be a polyfill issue.

We're working on trying to isolate the problem so that we can share a repro for this problem. In the meantime, we wanted to file this in case: a) anyone familiar with parseScript can guess what's going on, or b) other people writing single-page apps are hitting this problem and can offer additional insight.

JanMiksovsky:
I spent quite a bit of time trying to isolate a clean repro case for this problem outside the context of our application, but have been unable to do so. Nevertheless, we're still hitting this problem in our app. I'm happy to offer more information to anyone interested in looking at this in more detail.

For the time being, I've submitted a PR to work around the problem by defending against the cast where `script.parentNode` doesn't exist.

TimvdLippe:
Thanks for the issue. It has been a great while since this issue was opened. Sadly, this issue misses a live reproduction with the most recent version of the webcomponents polyfill for us to be able to debug and potentially fix. Versions 1 of the polyfills fix a large amount of issues, probably including this one as well.

Please see [the CONTRIBUTING.md](https://github.com/webcomponents/webcomponentsjs/blob/master/CONTRIBUTING.md) for guidelines on filing issues.  You can start from [this jsBin template](http://jsbin.com/birajez/edit?html,console,output) as a starting point and link it in the issue template under "Live Demo".

Feel free to file a new issue following these instructions if you can still reproduce this issue with versions 1 of the polyfills.  Thank you!

PR #343, opened 2015-07-07
Fix issue #342: check whether a script's parent exists before removing script from that parent

JanMiksovsky:
It's still unclear why this condition can arise. Nevertheless, if the condition does arise, and the script has no parent, then presumably it's no longer important to remove the script from its parent.


azakus:
Looks pretty innocuous, so I have no problem merging it.

PR #352, opened 2015-07-16
Fixes #351

### LINKED ISSUE #351: Custom Elements polyfill does not upgrade element with unused is attribute
Given that `x-foo` is registered as a non-type extension element and then the element is created like this:

```
<x-foo is="some-definition"></x-foo>
```

The element will not upgrade. It should become the element registered at `x-foo`. This is a weird thing to do but the syntax should be supported.

sorvell:
Fixes #351: : an element upgrades first via its tag name and only secondarily by its is attribute.


kevinpschaaf:
LGTM

PR #358, opened 2015-07-24
Fixes #354: Allow properties to be set on `Polymer` object: helps preserve compat…

### LINKED ISSUE #354: `Polymer` object should allow properties to be set
In Polymer 1.0, setting `dom = 'shadow'` on the `Polymer` object is the directive to enable Shadow DOM. The `webcomponents.js` polyfill currently defeats this command.

The full `webcomponents.js` polyfill defines a `Polymer` object to help facilitate calling `Polymer` when using Polymer version 0.5. Unfortunately, this feature poorly interacts with Polymer 1.0 by resetting the Polymer object. Having specific support for Polymer in the polyfill is not ideal, but this feature cannot be removed without breaking Polymer 0.5. It should, however, be made more forgiving to modifications of the Polymer object.

sorvell:
…ibility with Polymer 0.5 and 1.0

### DIFF
--- a/src/WebComponents/dom.js
+++ b/src/WebComponents/dom.js
@@ -63,6 +63,12 @@
     }
     elementDeclarations.push(arguments);
   };
+
+  // copy any properties off an existing Polymer object
+  if (window.Polymer) {
+    mixin(polymerStub, window.Polymer);
+  }
+
   window.Polymer = polymerStub;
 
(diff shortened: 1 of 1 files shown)


sorvell:
We can just remove this feature since Polymer 0.5 is pinned to an older webcomponents.js

PR #359, opened 2015-07-24
Fixes #355 install `innerHTML` on Template elements.

### LINKED ISSUE #355: Template polyfill should support innerHTML
The polyfill currently patches `document.createElement` to create a content fragment; it should also add an `innerHTML` accessor that populates `content`, the same as the native element.

sorvell:


src/Template/Template.js
```diff
@@ -32,6 +34,28 @@ if (typeof HTMLTemplateElement === 'undefined') {
       while (child = template.firstChild) {
         template.content.appendChild(child);
       }
+      // add innerHTML to template
+      Object.defineProperty(template, 'innerHTML', {
+        get: function() {
+          var o = '';
+          for (var e = this.content.firstChild; e; e = e.nextSibling) {
+            if (e) {
```
jmesserly: the loop condition `e` should make this check redundant?

src/Template/Template.js
```diff
@@ -32,6 +34,28 @@ if (typeof HTMLTemplateElement === 'undefined') {
...
+              o += e.outerHTML || e.textContent;
```
jmesserly: fwiw, I think the HTML spec has escaping for text nodes:
https://html.spec.whatwg.org/multipage/syntax.html#serialising-html-fragments
> If current node is a Text node:
> If the parent of current node is a style, script, xmp, iframe, noembed, noframes, or plaintext element, or if the parent of current node is a noscript element and scripting is enabled for the node, then append the value of current node's data IDL attribute literally.
> Otherwise, append the value of current node's data IDL attribute, escaped as described below.

In this case parent is template, so my guess is it should escape. Escape algorithm is:
> Escaping a string (for the purposes of the algorithm above) consists of running the following steps:
> 1. Replace any occurrence of the "&" character by the string `"&amp;"`.
> 2. Replace any occurrences of the U+00A0 NO-BREAK SPACE character by the string `"&nbsp;"`.
> 3. If the algorithm was invoked in the attribute mode, replace any occurrences of the """ character by the string `"&quot;"`.
> 4. If the algorithm was not invoked in the attribute mode, replace any occurrences of the "<" character by the string `"&lt;"`, and any occurrences of the ">" character by the string `"&gt;"`.

Here's an example:
```js
var t = document.createElement('template');
t.innerHTML = 'foo < bar > baz'
t.innerHTML
// prints: "foo &lt; bar &gt; baz"
t.firstChild.textContent
// prints: "foo < bar > baz"
```

And here's some code I wrote or ported from html5lib. (apologies for Dart instead of JS!)
https://github.com/dart-lang/html/blob/317a653cb7ff51c18ef3f632336e1ef72641cc3c/lib/dom_parsing.dart#L192

jmesserly: Ah, found a better example: ShadowDOM's innerHTML, from this very repository :)
https://github.com/webcomponents/webcomponentsjs/blob/938051272519053867b5bafa7411e7afefdd2c01/src/ShadowDOM/wrappers/HTMLElement.js#L34

sorvell: Sounds good. Thanks for catching the sloppiness.

src/Template/Template.js
```diff
@@ -61,5 +83,26 @@ if (typeof HTMLTemplateElement === 'undefined') {
...
+
+    function escapeReplace(c) {
+      switch (c) {
+        case '&':
+          return '&amp;';
+        case '<':
+          return '&lt;';
+        case '>':
+          return '&gt;';
+        case '"':
```
kevinpschaaf: Is this supposed to be in the regex?

jmesserly: I think what happened is it came from ShadowDOM code:
https://github.com/webcomponents/webcomponentsjs/blob/938051272519053867b5bafa7411e7afefdd2c01/src/ShadowDOM/wrappers/HTMLElement.js#L31

they have two regexs because there are two escape modes in the HTML spec, depending on if "attribute mode" is set.
https://html.spec.whatwg.org/multipage/syntax.html#serialising-html-fragments

jmesserly: > If the algorithm was invoked in the attribute mode, replace any occurrences of the """ character by the string "&quot;".

> If the algorithm was not invoked in the attribute mode, replace any occurrences of the "<" character by the string "&lt;", and any occurrences of the ">" character by the string "&gt;".

jmesserly:
LGTM! Could maybe remove the unused case for quotes as @kevinpschaaf suggested, but otherwise this is awesome.

kevinpschaaf:
LGTM

Issue #366, opened 2015-07-30
HTML Imports: Edge browser doesn't need fakeLoad for inline style elements


erikcharlebois:
I'm now getting load events on inline style elements in Edge, which is causing havoc on the import process cause the fakeLoad event and the real load event both fire. Quick hack to turn isIE11 into isIE11OrOlder fixed it for me.

Test from console in Edge:
what = document.createElement('style')
what.addEventListener('load', function() { console.log('loaded'); })
document.head.appendChild(what)

PR #367, opened 2015-08-01
Don't fake a load event for styles on Edge

rictic:
Fixes https://github.com/webcomponents/webcomponentsjs/issues/366 

Also fixes the following tests on Edge:

ShadowCss
  * html/style-import-base-tag.html

Document
  * document.registerElement attachedCallback, detatchedCallback
  * document.registerElement get reference, upgrade, rewrap


azakus:
LGTM

jmesserly:
awesome!

PR #372, opened 2015-08-06
Use basic ponyfill for requestAnimationFrame

rictic:
in CustomElements.js

Fixes an error in certain old (and probably technically
unsupported) Android browsers.

### DIFF
--- a/src/CustomElements/boot.js
+++ b/src/CustomElements/boot.js
@@ -78,13 +78,16 @@ function bootstrap() {
   // async to ensure *native* custom elements upgrade prior to this
   // DOMContentLoaded can fire before elements upgrade (e.g. when there's
   // an external script)
-  // Delay doubly to help workaround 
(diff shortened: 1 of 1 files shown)


jmesserly:
couldn't this be loaded as another polyfill code? I mean, it's really small, but in general, I think we try and keep polyfills separate. Seems out of scope for webcomponents.js ... just my opinion tho

jmesserly:
CC @sorvell @kevinpschaaf ... since the rAF dependency is newish, maybe they have thoughts.

notwaldorf:
@rictic ponyfill! 🐴😍

rictic:
I found a new word so I had to use it! http://kikobeats.com/polyfill-ponyfill-and-prollyfill/

sorvell:
LGTM seems ok due to minimal size...

PR #385, opened 2015-08-20
Rationalize our isIE variables.

rictic:
In CustomElements we were setting scope.isIE11OrOlder in boot.js rather than
base.js, such that it was undefined in register.js

Also switched to a more consistent naming scheme. isIE matches IE and
isIEOrEdge matches IE or Edge.

Fixes https://github.com/Polymer/polymer/issues/2310


sorvell:
LGTM

PR #387, opened 2015-08-21
Change import selectors to allow for imports with additional relation types

tuespetre:
I noticed in Chrome that the native imports functionality allows for additional relation types, but in browsers where the polyfill is used, those imports would not be processed. 

As a developer I want to be able to specify additional relation types for imports so that I can perform additional processing of certain imports using the established concept of relations (and extension relations) rather than `data-*` attributes.

### DIFF
--- a/src/CustomElements/traverse.js
+++ b/src/CustomElements/traverse.js
@@ -69,7 +69,7 @@ function _forDocumentTree(doc, cb, processingDocuments) {
     return;
   }
   processingDocuments.push(doc);
-  var imports = doc.querySelectorAll('link[rel=' + IMPORT_LINK_TYPE + ']');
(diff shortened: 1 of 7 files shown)


azakus:
Just to clarify, what other strings are you using for the `rel` attribute?

A few tests would be nice.

tuespetre:
What tests did you have in mind? I made sure the existing imports tests passed in browsers where the polyfill would take effect.

tuespetre:
Ah, I suppose I will whip up. A couple that verify for when there are multiple relations and in varying orders.

tuespetre:
A few tests ended up being more than nice; caught some things that I had missed and took the opportunity to dedupe some of the variables and functions regarding import selection. Please have another look.

tuespetre:
Just caught on that I might need to accept the CLA for this to go anywhere. I've done that now.

PR #389, opened 2015-08-23
Fix for re-insertion of element into the same place (to match native behavior)

### LINKED ISSUE #388: `attachedCallback` goes into infinite loop
Basically `this.parentNode.appendChild(this)` should not cause triggering `attachedCallback` (it doesn't in Chromium with native support).
Otherwise simple trick (using Polymer) causes infinite loop:

``` coffeescript
Polymer(
    'is' : 'my-element'
    attached : ->
        document.body.parentNode.appendChild(@)
)
```

nazar-pc:
Fixes #388

### DIFF
--- a/src/ShadowDOM/wrappers/Node.js
+++ b/src/ShadowDOM/wrappers/Node.js
@@ -367,6 +367,10 @@
   Node.prototype = Object.create(EventTarget.prototype);
   mixin(Node.prototype, {
     appendChild: function(childWrapper) {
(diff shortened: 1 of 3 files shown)


addyosmani:
Deferring to @azakus as he had questions about the need for this change in https://github.com/webcomponents/webcomponentsjs/issues/388#issuecomment-134333469

TimvdLippe:

PR #390, opened 2015-08-25
Use self instead of this to refer to window/context

ebidel:
R: @garlicnation 

This fixes an issue when running the wc.js polyfill through vulcanize + babel. The issue stems from the fact that Babel [redefines the top level `this` to `undefined`](https://github.com/babel/babel/commit/fb360039cea6c7e8e4f09458b518b1fcf45d0d74) to match module semantics. Due to this, if you try to vulcanize index.html and then run babeljs over the .js from cripser, you'll get this output:

```
  }
  scope.URL = jURL;
})(undefined);
```

The simple fix is to use `self` to refer to window instead of `this`. I'm [not a good reason `this` was used in the first place](https://github.com/Polymer/URL/commit/51ab442b462e3386b628a0bc135f4123f588d559#comments). `self` is a better choice to refer to `window`.

Technically, one shouldn't be using babel on an file not written for es6, but this fix is simple enough to make things work. I think we should do it to support users.

### DIFF
--- a/src/MutationObserver/MutationObserver.js
+++ b/src/MutationObserver/MutationObserver.js
@@ -565,4 +565,4 @@
     global.MutationObserver = JsMutationObserver;
 
 
-})(window);
+})(self);
--- a/src/URL/URL.js
+++ b/src/URL/URL.js
@@ -613,4 +613,4 @@
 
   scope.URL = jURL;
 
-})(this);
+})(self);


addyosmani:
`this` refers to the current context while `self` refers to the window. I agree `self` is a better reference for `window` per the above. Change LGTM, but deferring to @garlicnation for his review.

azakus:
Just curious, why doesn't a reference to `window` work?

ebidel:
It does, but I'd image some of the standalone libs are interesting in envs without `window`.

azakus:
Maybe, but lots of devs also use `self` as context reference in their own code. I wonder if it would break anything to use `self` here if webcomponents.js is bundled or vulcanized.

ebidel:
The original of assuming `this` is the global object it equally as bad IMO. If someone uses the library standalone, they're loading it in the global context. I can see this affecting node users that set `self = SOMETHING` in their script, but that's not our primary audience. I'd be surprised if many people redefine self in the global context.

azakus:
That's a fair point. I think this is a fine change in light of that. Maybe update this PR to swap *all* the window references to self?

addyosmani:
I think merging this once all references have been switched up and we've verified test passing status hasn't been impacted sgtm.

ebidel:
Hey guys. I don't feel comfortable switching all instances of `window` There's a lot of code in here I've never seen. How about we merge this PR as is and we can go back through later to possibly make that change?

azakus:
Created a tracking issue for that: #394.

PR LGTM, so I'll merge it.

addyosmani:
:+1:

PR #392, opened 2015-08-26
Add WebComponents.whenReady

azakus:
A callback based timing function equivalent to the WebComponentsReady event

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,17 @@ window.addEventListener('WebComponentsReady', function(e) {
 });
 ```
 
+### `WebComponents.whenReady`
+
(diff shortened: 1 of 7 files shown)


addyosmani:
Requires a rebase if we're still interested in landing this (cc @sorvell)

TimvdLippe:
This has been natively implemented as [`whenDefined`](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined) and supported in the polyfill: https://github.com/webcomponents/custom-elements/blob/6ad9993949a4f25c05b50b27a031a2ebfc49df89/src/CustomElementRegistry.js#L221

PR #393, opened 2015-08-28
Transforms Node.childNodes into a live NodeList

### LINKED ISSUE #321: Breaks directive resolution of Angular.js in Safari and Firefox.
**Update:** I've made this, so you can test it even faster (check in Chrome - should work, and FF and Safari fail): http://plnkr.co/edit/r2GE2x4vTPBZscMTrn2u?p=preview

**Without** the webcomponents.js script, the `<custom-container>` from this markup:
```html
<!DOCTYPE html>
<html lang="en"><head>
	<title>Repro angular/webcomponents.js problem</title>

	<!-- It's OK when not using webcomponents.js -->
	<!-- <script src="bower_components/webcomponentsjs/webcomponents.js"></script>-->

</head>
<body ng-app="app">

	<custom-container></custom-container>

<script src="bower_components/angular/angular.js"></script>
<script src="js/app.js"></script>
</body>
</html>
```
is resolved to:
```html
<div>
    <span>Resolved!</span>
</div>
```
Here is my angular code:
```javascript
var app = angular.module( "app", [
  "custom-container-module"
] );

angular.module( "custom-container-module", [
    "inner-container-module"
  ] )
  .directive( "customContainer", function() {
    return {
      restrict: "E",
      replace: true,
      template: "<div><inner-container></inner-container></div>"
    };
  } );

angular.module( "inner-container-module", [] )
  .directive( "innerContainer", function() {
    return {
      restrict: "E",
      replace: true,
      template: "<span>Resolved!</span>"
    };
  } );
```

**The problem:**
 - If I uncomment this:
```html
 <!-- <script src="bower_components/webcomponentsjs/webcomponents.js"></script> -->
```

Only the first-level template is resolved (**Safari & FF**, haven't tested in IE or Opera and of course Chrome is OK). So, I just get this:
```html
<div>
   <inner-container></inner-container>
</div>
```

I've been debugging Angular since yesterday and I see that the inner template is found, but there is something weird about adding the elements to element.childNodes - element.childNodes seems to be immutable in Safari and FF...

Any help would be appreciated. Thank you very much for your time and effort.

juliomelo:
Transforms `Node.childNodes` into a live `NodeList` - fixes #321 

According to [DOM Level 1 Specification](http://www.w3.org/TR/REC-DOM-Level-1/level-one-core.html), regarding to childNodes attribute:

> The content of the returned NodeList is 'live' in the sense that, for instance, changes to the children of the node object that it was created from are immediately reflected in the nodes returned by the NodeList accessors; it is not a static snapshot of the content of the node.

This changeset introduces a permanent reference for `childNodes` `NodeList` that is created on-demand. Once created, it's continuously updated on node operations, such as `appendChild`, `insertBefore`, `removeChild`, set `textContent` and `innerHTML`.

No performance issues were realized, during tests on Firefox 39. Without the fix, the tests runned in 28.670 seconds (mean of 5 consecutive tests). With the fix, 28.746 seconds.

### DIFF
--- a/src/ShadowDOM/wrappers/HTMLElement.js
+++ b/src/ShadowDOM/wrappers/HTMLElement.js
@@ -23,6 +23,7 @@
   var unwrap = scope.unwrap;
   var wrap = scope.wrap;
   var wrappers = scope.wrappers;
+  var rebuildChildNodes = scope.rebuildChildNodes;
 
   /////////////////////////////////////////////////////////////////////////////
   // innerHTML and outerHTML
@@ -204,6 +205,10 @@
         setInnerHTML(this.content, value);
       } else {
         unsafeUnwrap(this).innerHTML = value;
+
+        if (this.childNodes_ !== undefined) {
+          rebuildChildNodes(this);
+        }
       }
 
       var addedNodes = snapshotNodeList(this.childNodes);
--- a/src/ShadowDOM/wrappers/Node.js
+++ b/src/ShadowDOM/wrappers/Node.js
@@ -29,6 +29,11 @@
   var wrap = scope.wrap;
   var wrapIfNeeded = scope.wrapIfNeeded;
   var wrappers = scope.wrappers;
+  var insertBeforeNodeList = scope.insertBeforeNodeList;
+  var clearNodeList = scope.clearNodeList;
+  var indexOfNodeList = scope.indexOfNodeList;
+  var removeItemAtNodeList = scope.removeItemAtNodeList;
+  var copyNodeList = scope.copyNodeList;
 
   function assertIsNodeWrapper(node) {
     assert(node instanceof Node);
@@ -136,8 +141,12 @@
   }
 
   function snapshotNodeList(nodeList) {
-    // NodeLists are not live at the moment so just return the same object.
-    return nodeList;
+    var snapshot = new NodeList();
+    for (var i = 0; i < nodeList.length; i++) {
+      snapshot[i] = nodeList[i];
+    }
+    snapshot.length = i;
+    return snapshot;
   }
 
   // http://dom.spec.whatwg.org/#node-is-inserted
@@ -227,6 +236,9 @@
         childWrapper = nextSibling;
       }
       wrapper.firstChild_ = wrapper.lastChild_ = null;
+      if (wrapper.childNodes_ !== undefined) {
+        clearNodeList(wrapper.childNodes_);
+      }
     } else {
       var node = unwrap(wrapper);
       var child = node.firstChild;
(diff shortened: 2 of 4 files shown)


jmesserly:
This makes DOM insert/remove O(N) once childList is requested. It's not the kind of performance change that will show up in tests, you'll need a benchmark to know.

Yes, the specs say a live childList (although note: NodeList is an [old-style collection](https://dom.spec.whatwg.org/#old-style-collections)), but they also don't want insert/remove to be O(N). It's a choice of violating an uncommonly relied on behavior, or a commonly relied on performance expectation. We're stuck between a rock and a hard place. The original author of the polyfill (not me, fwiw) did a bunch of research, and came to the conclusion that the current implementation is the best trade off. So I wouldn't second guess it lightly.

That said, the one can make a perf argument in favor of this change: `get childNodes` is not supposed to be O(N) either. After this change, it becomes O(1). If code can always do mutations first before requesting childNodes, the overall cost is either the same or better.

My personal opinion, FWIW, is that this is probably worth doing. But it'd be nice to see more benchmarks, including what it does in the worst case, what impact is on angular and other frameworks, etc, to build a strong case on the performance side of things.

jmesserly:
CC @addyosmani and @azakus for thoughts.

jmesserly:
BTW, I wanted to add, it's super awesome you put this together. Very clean implementation too. :)

addyosmani:
This is a really interesting set of changes (and thanks for working on it, @juliomelo!). Per @jmesserly I do think we'll want to see more data around benchmarks before we can decide how to proceed. I'd be interested to see how this impacts raw polyfill use and use with Polymer if possible.

addyosmani:
Thanks for continuing to work on this. Any further thoughts on the benchmarking comments?

juliomelo:
Thanks for your analysis. I agree with all you about the importance of the benchmarks. Unfortunately my team decided to drop webcomponents from our project, and thus become difficult to dedicate time to this contribution.

addyosmani:
@jmesserly Given the above, we probably want to make a call on whether the potential perf benefits this brings are worth landing it vs. holding off for more time spent researching impact of the change. I won't have time this Q to look at the data closely, but the change otherwise LGTM. Do you or @azakus want to make a call on this?

parga:
@addyosmani @juliomelo will be this issue fixed? we have a components library created using web components, and we are using them inside angular, so this will be really helpful it gets solved, 

an example we have is using ng-repeat inside the web-component does not update the web component state in FF and Safari

freeman:
@addyosmani Any news on this issue ? We are in the same case as @parga

TimvdLippe:

morewry:
This still seems to be an issue. Is https://github.com/webcomponents/shadydom/pull/226 related?

Issue #329, opened 2015-06-06
base tag ignored for HTMLImports


garysoed:
E.g.:
```html
<base href="base">

<link rel="import" href="a/b/c.html">
```
The example above tries to load `a/b/c.html` instead of `base/a/b/c.html`

jmesserly:
@sorvell may have thoughts on this one?

fwiw, I don't see anything in [HTML Imports spec](http://www.w3.org/TR/html-imports/) specifically addressing the [base element](https://html.spec.whatwg.org/multipage/semantics.html#the-base-element), but I think it falls out of existing HTML spec language for [document-base-url](https://html.spec.whatwg.org/multipage/infrastructure.html#document-base-url).

I think the issue is that HTML Imports polyfill always creates a `<base>` element in attempt to get the normal URL resolution behavior, and adds it first in `<head>`
https://github.com/webcomponents/webcomponentsjs/blob/c15e80fedc2322c6ac0eccaa7e8501a23ee9f7af/src/HTMLImports/importer.js#L132

Maybe it should have some logic to only add that if there isn't a 'base' already?

I might be missing something obvious here though, not an expert on the imports code :)

drarmstr:
I ran into this problem also.  As motivation: I was using the <base> tag in a web component so I could use <a> links in the component with just a hash anchor path for a router.  Because the links are in a component that is not in the root directory, the browser would append the path to the component when following the link which is obviously incompatible with a router.  While a solution might have been to use "/#path" for the links this didn't work in my case because with "/#..." browsers also clear the current URL query string.  Using <base href="../../"/> in a component seems like a great solution to allow use of hash paths in links in components.

This does in fact work great in Chrome.  However, it seems that the <base> tag is not respected in this Polyfill creating a compatibility problem...

Example:
index.html:

    <!DOCTYPE html>
    <html lang="en">
    <head>
        <meta charset="utf-8" />
        <script src="components/webcomponentsjs/webcomponents-lite.js"></script>    
        <link rel="import" href="components/foo/foo.html" />
    </head>

    <body>
        Main body
    </body>
    </html>
    
components/foo/foo.html:

    <base href="../../" />
    <link rel="import" href="components/bar/bar.html" />
    <script>alert("foo");</script>
    <template>
        <a href="#path/for/router">link</a>
    </template>

components/bar/bar.html:

    <script>alert("bar");</script>

In Chrome the "bar" alert will show up properly, but it will not with the Polyfill.

[...]

PR #395, opened 2015-09-03
Properly consider <base> tags for HTML imports

drarmstr:
Discussed with @jmesserly for Issue #329 

I was added to the corporate-level Contributor License Agreement with Intel under douglas.armstrong@intel.com


jmesserly:
LGTM. @sorvell want to take a look as well?

If it's not too much trouble, might be good to add a unit test as well, that way we don't accidentally regress it.

Thanks for putting this together!

addyosmani:
LGTM2. Agree on us landing a test alongside this change. Think you might have time for that @drarmstr

addyosmani:
friendly ping @drarmstr

drarmstr:
I provided my local test case in #329.  I'm not too familiar with your test system how that would integrate.

drarmstr:
Is this pull request still a candidate for merging?

azakus:
This PR fails to complete test runs in firefox and safari.

src/HTMLImports/importer.js
```diff
@@ -128,22 +128,28 @@ function makeDocument(resource, url) {
...
-  // establish a relative path via <base>
-  var base = doc.createElement('base');
-  base.setAttribute('href', url);
-  // add baseURI support to browsers (IE) that lack it.
-  if (!doc.baseURI && !hasBaseURIAccessor(doc)) {
-    // Use defineProperty since Safari throws an exception when using assignment.
-    Object.defineProperty(doc, 'baseURI', {value:url});
+  
+  // install html
+  doc.body.innerHTML = resource;
```
azakus: why is this reordered before setting the `<base>` tag?

drarmstr: My thinking was so the doc content would be set so we could search for any existing `<base>` tags with the `doc.querySelector('base')` line below.  Is there a concern with this?

drarmstr:
@azakus Trying to keep this moving along I tried to investigate the travis-ci failure in case that is holding things up.  I setup a web-component-tester environment and tried to duplicate the tests.  I do see failures in the HTMLImports test on FireFox and Chrome.  However, I also see failures when I run with the unmodified webcomponentsjs master.  So, perhaps these are existing known issues?

TimvdLippe:
This was fixed in https://github.com/webcomponents/html-imports/commit/2b10492d48b8d5fdd4187510ee6175e0f25967ce Sorry for the late response!

Issue #180, opened 2015-01-29
Throwing error on flags.log.split()


miguelmota:
`flags.log` is an object

![screen shot 2015-01-29 at 1 22 52 pm](https://cloud.githubusercontent.com/assets/168240/5966529/28efc7c8-a7ba-11e4-8a88-9cccd855eb39.png)
![screen shot 2015-01-29 at 1 23 06 pm](https://cloud.githubusercontent.com/assets/168240/5966530/2900d8f6-a7ba-11e4-892b-8176ec6e669a.png)

miguelmota:
Fixed by moving script to head.

addyosmani:
Thanks for reporting back.

ComFreek:
This also happened to me. In my case I even had to move the `<script>` tag before all the `<link>` tags I had.

Dade3007:
I'm still having the problem. I put the webcomponents.js script the first of all, but still having the error.

nislowk:
also having this issue.  Went away in a recent version of webcomponents.js but came back

igorspasic:
Also having the issue.

MauriceSchoenmakers:
maybe but it is still an errer IMHO

garlicnation:
https://github.com/webcomponents/webcomponentsjs/pull/223

garlicnation:
Fixed in #223. Reopen after the next release(Tenatively march 15th) if this reoccurs.

dtracers:
is #223 in the 6.0 release?

Because I downloaded 6.0 via bower and manually but am still experiencing this error.

Specifically it is saying on line 30 of the webcomponentjs is not a function

```
if (flags.log) {
    var parts = flags.log.split(",");
```

This is what is in the file but I feel it is supposed to be

```
if (flags.log && flags.log.split) {
```

ballerabdude:
@dtracers this is what im doing and its currently working except this version injects script tags into the dom. With the above fix other developers wont be able to see your bug fix in your node_moudle because its generally ignored.  
```javascript
"webcomponentsjs": "git@github.com:webcomponents/webcomponentsjs.git#58c870917c838846fb7f4c9ec72941c2358562cc"

PR #398, opened 2015-09-15
Make flags.log check more expressive #180

Kureev:
## Bug story

Requiring `webcomponents.min.js` leads to having #180 error. As far as I understand, this issue has been fixed #223, but I think we can use more expressive way for this.
## What have been done

Replaced `flags.log && flags.log.split` by `isString(flags.log)`. There are to ideas behind this change:
- Express that `flags.log` should be a `String` to match this case
- Avoid possible mistakes by passing object that have `split` method, but not `String`

I hope you will consider this minor change meaningful.

P.S. Please, bump a bower package version with this fix.

### DIFF
--- a/webcomponents.js
+++ b/webcomponents.js
@@ -9,6 +9,10 @@
  */
 
 (function() {
+  // Helper function to check if given source is a string or not
+  function isString(source) {
+    return Object.prototype.toString.call(source) === '[object String]';
+  }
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Seems that #223 solves this sufficiently. We haven't observed this error any longer.

PR #404, opened 2015-09-25
Include rAF polyfill in wc-lite builds

addyosmani:
Address an issue where headless webkit instances may not have `window.requestAnimationFrame` and thus may cause scripts to stop executing when hitting this.

We currently already include dom.js (with a rAF polyfill) in the full wc.js builds. This change just ensures that it is also included in the lite builds too.

### DIFF
--- a/src/WebComponents/build-lite.json
+++ b/src/WebComponents/build-lite.json
@@ -6,5 +6,6 @@
   "../HTMLImports/build.json",
   "../CustomElements/build.json",
   "../Template/Template.js",
+  "dom.js",
   "unresolved.js"
 ]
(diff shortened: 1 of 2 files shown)


ebidel:
Can you assign reviewers to PRs?

ebidel:
LGTM

jmesserly:
LGTM too :+1:

addyosmani:
Thanks for the LGTMs folks! Would be good to get one more LGTM, perhaps from @azakus before we merge this in.

azakus:
LGTM3

addyosmani:
:tada: Thank you!

PR #405, opened 2015-09-28
Fix for broken CustomEvent shim in IE

azakus:
Broken in #339


jmesserly:
LGTM. Sorry about that!

PR #415, opened 2015-10-14
add check for log split before executing

filaraujo:
seems that this change was propagated to the boot.js but not to the boot-lite.js


azakus:
yep, thanks!

PR #417, opened 2015-10-16
Fixes #416

### LINKED ISSUE #416: Make throttling attachedCallback when using MutationObserver polyfill an option
Currently when the MutationObserver polyfill is used, attached callbacks are throttled using setTimeout in an attempt to avoid flakiness around DOM mutation events (which the polyfills relies on) which are thought to be related to bugs when IE is under heavy DOM operations, although we have failed to recreate the "flakiness" in recent time.

Since using setTimeout makes it very difficult to reason about the completion of async tasks (as opposed to async tasks using microtask timing, which makes it somewhat easier to reason about), we are planning to disable that behavior, but make the setTimeout throttling an option so it can be used in emergency.

sorvell:
Fixes #416: (1) don't allow JsMutationObserver object to be redefined and set an explicit flag that the polypill is in use; (2) CustomElements now only throttles attached  if the `wc-throttle-attached` flag is explicitly set.


kevinpschaaf:
LGTM

PR #428, opened 2015-10-28
avoid TypeError on Firefox

valtido:
`TypeError: invalid 'instanceof' operand scope.wrappers.Window` error was showing on firefox, the `Window` object was undefined.


TimvdLippe:
This statement appears to be removed in https://github.com/webcomponents/shadydom Sorry for the late response!

PR #433, opened 2015-10-30
Ignore "async" imports for HTMLImports.whenReady

### LINKED ISSUE #431: HTMLImports.whenReady should not include `async` imports
HTMLImports.whenReady should not block on async `<link rel="import" async>` imports, to mach the native non-blocking behavior of HTML imports.

Solution may simply invlove adding `:not([async])` here: https://github.com/webcomponents/webcomponentsjs/blob/master/src/HTMLImports/base.js#L117

azakus:
Fixes #431

### DIFF
--- a/tests/HTMLImports/html/whenready.html
+++ b/tests/HTMLImports/html/whenready.html
@@ -1,0 +1,44 @@
+<!DOCTYPE html>
+<!--
+    @license
+    Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+    This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+    The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+    The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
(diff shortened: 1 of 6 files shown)


tests/HTMLImports/html/whenready.html
```diff
@@ -0,0 +1,44 @@
...
+  <script src="../../../../web-component-tester/browser.js"></script>
+  <link rel="import" href="imports/async-sync.html">
+  <link rel="import" href="imports/async-async.html">
+</head>
+<body>
+  <script>
+  test('whenReady ignores async', function(done) {
+    HTMLImports.whenReady(function (detail) {
+      assert(detail.loadedImports, 1, 'whenReady should wait for the sync import');
+      done();
```
kevinpschaaf: How about also add `assert.equal(syncLoaded, 1, 'sync load count incorrect');` to ensure the 1 loaded import was in fact the sync import?

azakus: `syncLoaded` is only incremented by `async-sync.html`

PR #434, opened 2015-11-04
Fix event.defaultPrevented for IE

azakus:
Only set defaultPrevented if event is cancelable
Add tests with and without cancelable
Fix shims in HTMLImports and CustomElements to support cancelable


azakus:
@sorvell @kevinpschaaf PTAL

sorvell:
LGTM

PR #435, opened 2015-11-05
Dedupe CustomEvent and Event shim constructor

azakus:
Moved to WebComponents/dom.js
Load dom.js for CE and HTMLImports polyfills
Add tests for functions in dom.js

General Test cleanup
- Remove old tests.js files
- Replace broken `?wc-shadow` for `?shadow=native`


sorvell:
LGTM :+1:

Issue #430, opened 2015-10-30
Support slot-based Shadow DOM v1


laco0416:
Hi teams.
Are you planning to support Shadow DOM v1?
(Spec: http://w3c.github.io/webcomponents/spec/shadow/)

ebidel:
Definitely! We've discussed starting this work this quarter or the next. @tjsavage for timelines.

tomalec:
+1 for adding v1 support :)

It is mentioned in [Polymer's 2016 Roadmap](https://blog.polymer-project.org/announcements/2016/02/16/2016-roadmap/)
It would be cool to see such roadmap for webcomponents.js as well.

Just to be able plan development, and decide according to which spec version should we code. Especially in regards of vanilla Web Components.

ebidel:
I think webcomponents.js support is implicit in supporting v1 in Polymer. cc @tjsavage @azakus @addyosmani

tomalec:
Once v1 will be shipped, will you still polyfill v0 more or less as Hayato proposed for Blink? https://github.com/w3c/webcomponents/blob/6100444d54d8b759f456fb56fb263473494bf28c/proposals/shadow-dom-v1-in-blink.md

Will you use v0 in Chome to polyfill v1, or are you waiting for v1 to be complete in there?

tomalec:
Is there any update on this? 

Implementation seems to be done in Chromium 
https://bugs.chromium.org/p/chromium/issues/detail?id=531990#c127
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/zrZRD2ls5tw/ibbjD3cQAQAJ

PR #437, opened 2015-11-06
polyfill support for slot element and ::slotted

uniqname:
Begins to address #430.

This solution appears to work fine for the polyfill, but slot would still not be supported for browsers that support ShadowDOM but not `<slot>`. I'd be very happy to append this if I could get some guidance on where best to branch the logic.

### DIFF
--- a/src/ShadowCSS/ShadowCSS.js
+++ b/src/ShadowCSS/ShadowCSS.js
@@ -243,6 +243,14 @@ var ShadowCSS = {
    *
    * scopeName menu-item {
    *
+   * And this rule:
+   *
+   * polyfill-next-selector { slotted: ':host nav-item'; }
+   * ::slotted(nav-item) {
+   *
+   * to this:
+   *
+   * scopeName nav-item {
   **/
   insertPolyfillDirectivesInCssText: function(cssText) {
     // TODO(sorvell): remove either content or comment
@@ -387,12 +395,16 @@ var ShadowCSS = {
   },
   /*
    * Convert combinators like ::shadow and pseudo-elements like ::content
-   * by replacing with space.
+   * and ::slotted by replacing with space.
   */
   convertShadowDOMSelectors: function(cssText) {
     for (var i=0; i < shadowDOMSelectorsRe.length; i++) {
       cssText = cssText.replace(shadowDOMSelectorsRe[i], ' ');
     }
+    for (var i=0; i < shadowDOMSelectorFunctionsRe.length; i++) {
+      cssText = cssText.replace(shadowDOMSelectorFunctionsRe[i], '$1 ')
+    }
+
     return cssText;
   },
   // change a selector like 'div' to 'name div'
@@ -587,7 +599,10 @@ var selectorRe = /([^{]*)({[\s\S]*?})/gim,
       /\/shadow-deep\//g, // former /deep/
       /\^\^/g,     // former /shadow/
       /\^/g        // former /shadow-deep/
-    ];
+  ],
+  shadowDOMSelectorFunctionsRe = [
+      /::slotted\((.*)\)/g,
+  ];
 
 function stylesToCssText(styles, preserveComments) {
   var cssText = '';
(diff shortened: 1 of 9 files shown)


azakus:
Wow, thanks for the starting the ball rolling!

Can you add some tests?
Some small ones that test the API for `name`, and some HTML tests that show distribution happened as expected.

I'll need some time to review the spec and see if there are any obvious gaps in your implementation.

@jmesserly if you have some time, it would be awesome to get more eyeballs on this.

uniqname:
Love to add some tests. I started looking at some of the existing ones as examples but didn't get too far. web-component-tester is new for me so I need to wrap my head around it.

azakus:
Here's the `<content>` tests in ShadowDOM polyfill as an example: https://github.com/webcomponents/webcomponentsjs/blob/master/tests/ShadowDOM/js/HTMLContentElement.js

web-component-tester's runtime is basically just mocha and chai with some tooling to play nicely with HTMLImports and the CLI webserver / test result printer.

uniqname:
I'm unclear on some details with `<slot>`. With the `content` element there was a distinction between what `<content select="*">` would select and what `<content>` would select. It's my understanding that the former selected all `ElementNode`s (`shadowHost.children`) whereas the latter selected all `Node`s (`shadowHost.childNodes`). It's unclear to me what `<slot>` should select (there is nothing like `<slot name="*">`). I assume `<slot>` is works as `<content>`, assigning _all_ nodes without an otherwise matching named slot, but the spec is unclear to me on this.

uniqname:
Updated with tests for `<slot>` but I don't have a way to validate if they pass. I've not been able to get any of the existing tests to pass, nor have any passed on travis. :/

azakus:
Yeah, the tests are in a funky state, and have been for a while. I plan to clean them up, but haven't found a good opportunity yet :P

uniqname:
/me wipes brow
glad it's not just me then :)

jmesserly:
yeah I'd be happy to take a look. I haven't fully loaded the new spec in my brain but it's really neat to see an implementation!

So one high level question -- are we trying to keep the old support too? It seems like it will be difficult to keep a superimposition of old-and-new ShadowDOM specs. Could we just support the V1 spec, in which case, distribution can be radically simplified right? There's no `<content>` anymore?

http://w3c.github.io/webcomponents/spec/shadow/

if I were tackling this, I'd be tempted to start by deleting the existing distribution algorithm (nuke it from orbit, only way to be sure! ) & writing the new one that followed the new spec, so the correspondence is easy to see.

uniqname:
For whatever it's worth, `<slot>` and `<content>` could exists side-by-side as the distribution algorithms are identical(?) once a correlation has been made between a distributor (`<slot>`, `<content>`) and the nodes they are projecting (identified via `content[select]` or `slot[name] + [slot]` attributes). Though it might be more trouble than it's worth. simpler === better :)

addyosmani:
I would stringly be in favour of a major release that just supported V1, however if it was possible to avoid breakage given similarity of the code paths involved I would also be supportive of that - if it can be done cleanly. Simpler is most definitely better. The only issue I can see is we need to balance avoiding breakage of existing apps vs. carrying backage that we otherwise may want to remove at some point. 

Anyone else have thoughts here?

nazar-pc:
I'd support approach where both `<content>` from v0 and `<slot>` from v1 are supported at the same time, but have completely separate implementations. So that v0 removal will be easier in future (probably even add some build option to exclude v0 stuff for those who don't need them).

The point is that many users of polyfill might already use it according to v0 spec. Giving ability to migrate elements one-by-one will make transition much less painful. AFAIK, Chrome does transition in the same way - implement v1 alongside with keeping v0 where it doesn't break the entire thing to give people some time to migrate.

lukasoppermann:
Hey, is this still needed? Because I currently have the problem that my selector `::slotted([column~="3l"])` does not work in firefox, while it works fine in Safari and Chrome.

Pilatch:
Please, for the love of Pete, can we get this merged if it actually works?

justinfagnani:
@Pilatch this PR is extremely old. ShadyDOM here supports `<slot>`: https://github.com/webcomponents/shadydom

@azakus can we close this?

TimvdLippe:
@justinfagnani Yup closing.

PR #439, opened 2015-11-10
Force templates to upgrade before elements

garlicnation:


garlicnation:
@sorvell @kevinpschaaf PTAL

azakus:
LGTM

rubenstolk:
Will we get a version increase for this?

PR #440, opened 2015-11-10
Missed upgradeDocumentTree

garlicnation:


sorvell:
The code change here is to `upgradeDocument` not `upgradeDocumentTree`. Can you explain why we need to ensure template upgrades when these are called? Just for safety or was there a specific use case?

garlicnation:
CustomElements bootstrapping kicks off upgradeDocumentTree, which was resulting in un-prepped templates(I thought). Turns out this doesn't fix the problem, though, so my identification of a cause was incorrect.

Issue #241, opened 2015-03-16
content doesn't refresh when attribute change


pienikrz:
I have a simple component

```
<polymer-element name="test-component">
    <template>
        <content select="div[order='1']"></content>
        <content select="div[order='2']"></content>
        <content select="div[order='3']"></content>
        <content select="div[order='4']"></content>
        <content select="div[order='5']"></content>
        <content select="div[order='6']"></content>
    </template>
    <script>
    Polymer('test-component', {
    });
    </script>
</polymer-element>
```

which I use on a simple page:

```
<!DOCTYPE html>
<html>
    <head>
        <meta http-equiv="content-type" content="text/html;charset=utf-8">
        <script src="webcomponents/webcomponents.js"></script>
        <script src="polymer/polymer.js"></script>
        <link href="test-component.html" rel="import">
    </head>
    <style>
    </style>
    <body>
        <test-component>
            <div order="1" onclick="this.setAttribute('order', '4');"> Div 1</div>
            <div order="2"> Div 2</div>
            <div order="3"> Div 3</div>
            <div order="4"> Div 4</div>
            <div order="5"> Div 5</div>
            <div order="6"> Div 6</div>
        </test-component>
    </body>
</html>
```

When I click on first div, it's order attribute change to "4" in all browser but only in Chrome the output is changing from:

```
Div 1
Div 2
Div 3
Div 4
Div 5
Div 6
```

to

```
Div 2
Div 3
Div 1
Div 4
Div 5
Div 6
```

adko-pl:
I have the same problem, @addyosmani  any estimates when this will be fixed?

jmesserly:
fwiw, the code that's supposed to handle this:
https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/ShadowRenderer.js#L611
used from: https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/wrappers/Element.js#L92

maybe it's failing to match that selector?
https://github.com/webcomponents/webcomponentsjs/blob/master/src/ShadowDOM/ShadowRenderer.js#L487

jmesserly:
hmm, when I ran that query on your selector, it seemed to work.
What might be happening: onclick="this.setAttribute('order', '4');" is giving you an unwrapped ShadowDOM node. That won't work, and we can't fix it in polyfill. Try onclick="ShadowDOMPolyfill.wrap(this).setAttribute('order', '4');"

[...]

PR #449, opened 2015-11-18
Fix for #241 - content doesn't refresh when attribute change

pienikrz:


TimvdLippe:

PR #457, opened 2015-12-02
Fixes #110: Adds support for Document/ShadowRoot activeElement.

### LINKED ISSUE #110: support for ShadowRoot/Document activeElement

bicknellr:


tests/ShadowDOM/html/activeElement.html
```diff
@@ -0,0 +1,344 @@
...
+      r_1_l = r_1.querySelector('[shadow-template-id=x-shadow-host-root-1-light]');
+      r_1_0 = r_1.shadowRoot.querySelector('[shadow-template-id=x-shadow-host-root-1-0]');
+        r_1_0_l = r_1_0.querySelector('[shadow-template-id=x-shadow-host-root-1-0-light]');
+      r_1_1 = r_1.shadowRoot.querySelector('[shadow-template-id=x-shadow-host-root-1-1]');
+        r_1_1_l = r_1_1.querySelector('[shadow-template-id=x-shadow-host-root-1-1-light]');
+
+    document.body.appendChild(r);
+
+    // Nodes need to be in the document to be focusable.
+    ShadowDOMPolyfill.renderAllPending();
```
bicknellr: Should I add something to the focus wrapper to call this before calling native focus? Without this, the nodes I'm trying to focus aren't actually in the document and don't get focused.

tests/ShadowDOM/html/activeElement.html
```diff
@@ -0,0 +1,344 @@
...
+    assert.equal(r.shadowRoot.activeElement, r_1);
+    assert.equal(r_0.shadowRoot.activeElement, null);
+    assert.equal(r_1.shadowRoot.activeElement, r_1_1_l);
+  });
+
+  test('focused element in different document', function() {
+    var docType = doc.implementation.createDocumentType(
+      'html', '-//W3C//DTD XHTML 1.0 Transitional//EN',
+      'http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd');
+    var otherDocument = doc.implementation.createDocument(
```
azakus: Use `doc.implementation.createHTMLDocument()`

azakus: Actually, this whole test shouldn't matter. `document.activeElement` shouldn't change if you manually focus an element in a different document.

tests/ShadowDOM/html/activeElement.html
```diff
@@ -0,0 +1,344 @@
...
+    document.body.appendChild(r);
+
+    // Nodes need to be in the document to be focusable.
+    ShadowDOMPolyfill.renderAllPending();
+  });
+
+  teardown(function() {
+    document.body.removeChild(r);
+
+    assert.equal(r_0_0.shadowRoot.activeElement, null);
```
bicknellr: These assertions don't work because the elements aren't attached to the document after the `removeChild` call above - I'll move it to the end. Also, are assertions within the post-test hook ok?

tests/ShadowDOM/html/activeElement.html
bicknellr: Removed.

bicknellr:
@azakus I think the tests that are failing at this point are unrelated to this PR.

tests/ShadowDOM/html/activeElement.html
addyosmani: Imo, it's worth doing so, but deferring to @azakus for a call here.

azakus: I'd say let's do it. If it's a performance problem, we can change it to living inside any of the focus-querying methods.

azakus:
LGTM, with move of `renderAllPending` into focus handling.

tests/ShadowDOM/html/activeElement.html
bicknellr: Done; conveniently there was already a list of other methods needing the same thing.

azakus:
LGTM, merging.

PR #463, opened 2015-12-12
Fix `activeElement` for ShadowRoot: focused elements in light DOM of host are now returned.

bicknellr:


bicknellr:
This doesn't handle the case that the active element is within the shadow DOM of an element in the ShadowRoot's host's light DOM.

bicknellr:
Ok, fixed for that case.

bicknellr:
@azakus ping

src/ShadowDOM/wrappers/ShadowRoot.js
```diff
@@ -80,8 +80,13 @@
 
       var activeElement = wrap(unwrappedActiveElement);
 
-      // Loop while activeElement is not a shallow child of this ShadowRoot.
-      while (!this.contains(activeElement)) {
+      if (activeElement === this.host) {
```
azakus: Please add a comment about how the host will not be considered the "activeElement" of a shadowroot.

azakus:
LGTM with nit

src/ShadowDOM/wrappers/ShadowRoot.js
bicknellr: Added.

bicknellr:
@azakus I think the failing tests are unrelated again - are these still expected for the time being?

azakus:
LGTM!

PR #469, opened 2015-12-22
Delete dead code.

mixed:
I think that It is dead code.


TimvdLippe:
This appears to be fixed in the mean time in https://github.com/webcomponents/shadydom Sorry for the late response!

PR #500, opened 2016-02-29
Set HTMLTemplateElement as a window property

### LINKED ISSUE #496: IE >= 10 breaks when in strict mode
Hi,

There seems to be an issue when importing webcomponents-lite.js from a file that declares "use strict" (getting "Variable undefined in strict mode"). Looks like this is coming from the following line (1066 in v0.7.21): 

`  if (needsTemplate) {
    HTMLTemplateElement = TemplateImpl;
  }`

My assumption is that the issue can be fixed if TemplateImpl will be assigned to "window.HTMLTemplateElement" instead?

MadaraUchiha:
Rather than as a global variable, which seems to have troubles with strict mode in a few browsers.

Closes #496

### DIFF
--- a/src/Template/Template.js
+++ b/src/Template/Template.js
@@ -225,7 +225,7 @@
   }
 
   if (needsTemplate) {
-    HTMLTemplateElement = TemplateImpl;
+    window.HTMLTemplateElement = TemplateImpl;
   }
 
 })();


azakus:
LGTM

MadaraUchiha:
Cheers!

Issue #497, opened 2016-02-24
Error in TemplateImpl.fixClonedDom in IE11


rictic:
Seeing this in a project's tests on attempting to update from v0.7.20 to v0.7.21

`t.parentNode` is null or undefined, leading to an error when we try to call `replaceChild` on it in the last line of `fixClonedDom`:

``` javascript
    TemplateImpl.fixClonedDom = function(clone, source) {
      // these two lists should be coincident
      var s$ = source.querySelectorAll(TEMPLATE_TAG);
      var t$ = clone.querySelectorAll(TEMPLATE_TAG);
      for (var i=0, l=t$.length, t, s; i<l; i++) {
        s = s$[i];
        t = t$[i];
        if (this.decorate) {
          this.decorate(s);
        }
        t.parentNode.replaceChild(s.cloneNode(true), t);
      }
    };
```

Here: https://github.com/webcomponents/webcomponentsjs/blob/84c6c670d1fb95ea848995339d952b96c2e8fe6a/src/Template/Template.js#L185

sorvell:
It'd be great to get a snippet of the template that's causing the error. Also, is this browser specific? (for example is it limited to IE?)

kapouer:
I'm seeing an error in fixClonedDom too, but sooner, on "source", which is sometimes a text node,
and which does not support querySelectorAll as such.
It's triggered by `myTextNode.cloneNode(true)` AFAICT.

kapouer:
Also, reverting to version 0.7.20 fixes the issue.

robbecker-wf:
I am getting this error only in IE11 (and not IE10!)
![image](https://cloud.githubusercontent.com/assets/6053699/13858516/056a9cd2-ec4d-11e5-9c74-9b6d72903a2f.png)
I've narrowed it down to when there is a template tag in the template for a custom element. It can be any template tag (bare, dom-if, dom-bind). Here's the simplest snippet that reproduces it for me in IE11:
```
<dom-module id="double-template">
    <template>
        <template>
            <div>hi</div>
        </template>
    </template>
    <script>
        Polymer({is: 'double-template'});
    </script>
</dom-module>
```
If you comment out the inner template, then no error.

robbecker-wf:
perhaps there is an async problem here when the template is stamped into the local DOM ?

PaulHMason:
[...]

PR #501, opened 2016-03-03
fixClonedDom: return if source is not an element

kapouer:


kapouer:
Might fix https://github.com/webcomponents/webcomponentsjs/issues/497 unless that issue is two-fold.

azakus:
LGTM

PR #507, opened 2016-03-10
Ensures extended element names are case-insensitive during registration

### LINKED ISSUE #506: Extended element name must be lowercase during registration
My understanding is that element names are not supposed to be case-sensitive. However, the current polyfill implementation fails to correctly upgrade extended elements if their registration doesn't specify the extended element type name in lowercase. The native implementations in Chrome and Firefox both handle this case successfully.

For example, the following code will display a log message under Chrome and Firefox's native implementations, but no message is displayed using the polyfill in Firefox:

```html
<!doctype html><body><script>
var prototype = Object.create(HTMLButtonElement.prototype);
prototype.createdCallback = function() { console.log("Created"); };
document.registerElement('examp-el', { extends: 'BUTTON', prototype: prototype });

document.body.appendChild(document.createElement('button', 'examp-el'));
</script>
```

jre-g:
Fixes #506 and updates a test case to catch regressions.

### DIFF
--- a/src/CustomElements/register.js
+++ b/src/CustomElements/register.js
@@ -91,6 +91,10 @@ function register(name, options) {
   }
   // record name
   definition.__name = name.toLowerCase();
+  // ensure extended name is also treated case-insensitively
+  if (definition.extends) {
+    definition.extends = definition.extends.toLowerCase();
+  }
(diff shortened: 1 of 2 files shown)


azakus:
LGTM

PR #514, opened 2016-03-22
Custom Elements v1

justinfagnani:


jogibear9988:
Any news on when this will be merged? Is Chrome also already supporting v1 of WebComponents? Any Polymer?

justinfagnani:
@jogibear9988 This is a bit of a longer/medium term effort. The spec hasn't fully settled, Polymer isn't ready to use v1, and Chrome isn't shipping an implementation yet.

trusktr:
Are any parts of the v1 API polyfill still TODO (besides possible changes needed after the API has settled)?

CarmenPopoviciu:
Do you have any idea what points of the spec still need to be settled? I've been following on the changes and I thought the APIs are pretty stable by now and that a consensus was reached

trusktr:
@justinfagnani IMHO, would be nice to get rid of the gulp build workflow, and just publish module files (letting the end user decide what build workflow to use). It would make the library more flexible. And idea: you could write ES6 modules in the `src` folder, then publish the files compiled to CommonJS in `dist` and have package.json's `main` field point to the dist entry point. The advantage of this is that advanced users don't have to use dist, and can use ES6 source files in their own build systems.

As it stands right now, working with the source files is difficult because they are chunks and the files cannot be used individually without having to manually load files in specific order. ES6 modules (and the CommonJS dist modules) solves that.

Additionally, you can publish a `global.js` file in the `dist` folder too, which basically can be a file that imports the main file, and attaches any required globals to the `window` (directly, or indirectly by importing other modules). This way `<script>` tag users can also have an easy way to use the library without any build system.

justinfagnani:
@rictic PTAL :)

rictic:
Is the travis failure unrelated? Looks like apt-get is failing to install?

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+    return node.nodeType === Node.ELEMENT_NODE
+  }
+
+  /**
+   * A registry of custom element definitions.
+   *
+   * See https://html.spec.whatwg.org/multipage/scripting.html#customelementsregistry
+   *
+   * @constructor
+   * @property {boolean} polyfilled Whether this registry is polyfilled
```
rictic: Recommend moving these down to the property assignments inside the constructor.

src/CustomElements/v1/README.md
```diff
@@ -0,0 +1,139 @@
+# Custom Elements v1
+
+## Status
+
+The polyfill should be mostly feature complete now. It supports defining
+custom elements, the custom element reactions, upgrades, and integrates with
+native Shadow DOM v1, and native and polyfilled HTML Imports.
```
rictic: This sentence stumbles a bit. Consider:

It supports defining custom elements, lifecycle reactions (e.g. attach, detach), and upgrading existing elements. It integrates with native Shadow DOM v1, as well as both native and polyfilled HTML Imports.

src/CustomElements/v1/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+The HTMLElement constructor is also specified to look up the tag name associated with a constructor by using `new.target`. `new.target` isn't available in ES5, is not polyfillable, but `this.constructor` behaves similarly for most ES5 class patterns, including the output of Babel, TypeScript and Closure, so the polyfill uses that to look up tag names.
+
+`new.target` isn't even feature detectable, since it's a syntax error in ES5. Because of this, the polyfill can't check `new.target` first and fallback to `this.constructor`. This also means that ES5-style constructors can't conditionally make a "super" call to the HTMLElement constructor (with `Reflect.construct`) in non-ES6 environments to be compatible with native Custom Elements.
+
+To allow for elements that work in both ES5 and ES6 environments, we provide a shim that overrides the HTMLElement constructor and calls `Reflect.construct` with either the value of `new.target` or `this.constructor`. This shim can only be executed in ES6 environments.
```
rictic: How would someone use this shim?

src/CustomElements/v1/README.md
```diff
@@ -0,0 +1,139 @@
...
+The Custom Elements V1 polyfill does not use the same module or build system as
+the other webcomponentsjs polyfills, and its build is not integrated into the default build tasks yet.
+
+To build run:
+
+    gulp CustomElementsV1
+
+This creates a CustomElementsV1.min.js file in dist/
+
+## Custom Elements in the DOM Spec
```
rictic: Are these notes only useful for people hacking on this code and reviewing it? An sentence introducing this part and the parter after would be helpful.

gulpfile.js
```diff
@@ -129,7 +131,21 @@ defineBuildTask('HTMLImports');
...
 
-gulp.task('build', ['webcomponents', 'webcomponents-lite', 'CustomElements', 
+gulp.task('CustomElementsV1', function () {
+  return gulp.src('./src/CustomElements/v1/CustomElements.js', {base: './'})
+      .pipe(closureCompiler({
+          compilation_level: 'ADVANCED',
+          warning_level: 'VERBOSE',
+          language_in: 'ECMASCRIPT6_STRICT',
+          language_out: 'ECMASCRIPT5_STRICT',
+          output_wrapper: '(function(){\n%output%\n}).call(this)',
```
rictic: With this you've got a two-layered wrapper on the output. Is that necessary? Consider just eliding this here.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+      // not a function, and always get observedAttributes.
+
+      // 7, 9.1:
+      var attributeChangedCallback = getCallback('attributeChangedCallback');
+
+      // 8, 9.2, 9.3:
+      var observedAttributes = constructor['observedAttributes'] || [];
+
+      // 15:
+      // @type {CustomElementDefinition}
```
rictic: Needs to be in a `/** */` to have any teeth.

gulpfile.js
```diff
@@ -129,7 +131,21 @@ defineBuildTask('HTMLImports');
...
+          externs: 'externs/html5.js',
+          js_output_file: 'CustomElementsV1.min.js'
+        }))
+      .pipe(gulp.dest('./dist'));
+});
```
rictic: Consider also running eslint over this file as well.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+ *  disconnectedCallback: Function,
+ *  attributeChangedCallback: Function,
+ *  observedAttributes: Array<string>,
+ * }}
+ */
+var CustomElementDefinition;
+
+(function() {
+  'use strict';
+
```
rictic: You probably want to detect a native implementation and early exit if so. Maybe also have a flag where you can force the polyfill... maybe.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+      }
+    },
+
+    /**
+     * @private
+     */
+    _unobserveRoot: function(root) {
+      if (root.__observer) {
+        root.__observer.disconnect();
+        root.__observer = null;
```
rictic: Looks like we should null out `__observer` after the body of the if statement.

rictic:
Added some changes here: https://github.com/rictic/webcomponentsjs/commit/f2c76cdf748bfe755e3d7c4b195f0ff9bc5176b8

Got to run, bbl.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+            // is observed?
+            this._addNodes(node.shadowRoot.childNodes);
+          }
+          if (node.tagName === 'LINK') {
+            var onLoad = (function() {
+              var link = node;
+              return function() {
+                link.removeEventListener('load', onLoad);
+                this._observeRoot(link.import);
+                this._addNodes(link.import.childNodes);
```
rictic: Are `link.import` or `link.import.childNodes` ever null? What about when the link element isn't `rel="import"`?

src/CustomElements/v1/CustomElements.js
sorvell: +1 the v0 polyfill also did this.

gulpfile.js
```diff
@@ -129,7 +131,21 @@ defineBuildTask('HTMLImports');
 defineBuildTask('ShadowDOM');
 defineBuildTask('MutationObserver');
 
-gulp.task('build', ['webcomponents', 'webcomponents-lite', 'CustomElements', 
+gulp.task('CustomElementsV1', function () {
+  return gulp.src('./src/CustomElements/v1/CustomElements.js', {base: './'})
+      .pipe(closureCompiler({
```
sorvell: Why use the `closureCompiler` here? None of the rest of this repo uses it. It seems better to consistently use it or not.

gulpfile.js
```diff
@@ -129,7 +131,21 @@ defineBuildTask('HTMLImports');
...
+          language_in: 'ECMASCRIPT6_STRICT',
+          language_out: 'ECMASCRIPT5_STRICT',
+          output_wrapper: '(function(){\n%output%\n}).call(this)',
+          externs: 'externs/html5.js',
+          js_output_file: 'CustomElementsV1.min.js'
+        }))
+      .pipe(gulp.dest('./dist'));
+});
+
+gulp.task('build', ['webcomponents', 'webcomponents-lite', 'CustomElements',
```
sorvell: Let's make a build of `webcomponents-lite` that includes the v1 polyfill, perhaps `webcomponents-v1-lite`? I think it's good to maintain the v0 version for awhile.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+
+  function createTreeWalker(root) {
+    // IE 11 requires the third and fourth arguments be present. If the third
+    // arg is null, it applies the default behaviour. However IE also requires
+    // the fourth argument be present even though the other browsers ignore it.
+    return doc.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, null, false);
+  }
+
+  function isElement(node) {
+    return node.nodeType === Node.ELEMENT_NODE
```
sorvell: `;`

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+    /** @private {Map<string, CustomElementDefinition>} **/
+    this._definitions = new Map();
+
+    /** @private {Map<Function, CustomElementDefinition>} **/
+    this._constructors = new Map();
+
+    this._whenDefinedMap = new Map();
+
+    /** @private {Set<MutationObserver>} **/
+    this._observers = new Set();
```
sorvell: Consider if prototype patching is a viable alternative to observation. The benefit would be timing consistency with native Custom Elements for dynamic operations. For example, these methods could be patched: `Element.prototype.appendChild`, `Element.prototype.removeChild`, `Element.prototype.insertBefore`, `Element.prototype.remove`, `Element.prototype.replaceChild`. The `innerHTML` accessor can also be patched (at least on all current browsers) as demonstrated here: http://jsbin.com/sihaqad/1/edit?html,output.

sorvell: This might also make it possible to implement the reactions queue as defined in the spec.

For clarity, with the patching approach, you'd miss parser generated element at page load time. A one time fixup would be needed to upgrade these.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+    this._attributeObserver =
+        new MutationObserver(this._handleAttributeChange.bind(this));
+
+    /** @private {HTMLElement} **/
+    this._newInstance = null;
+
+    this.polyfilled = true;
+    this.enableFlush = false;
+
+    this._observeRoot(document);
```
sorvell: Is this the only way elements in the document are upgraded? If so, this means the polyfill must be loaded before any elements in the main document are created. That's probably not acceptable. A one time fixup at load and/or (readyState == 'interactive') time could address this.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+      // 6, 7:
+      if (this._constructors.has(constructor)) {
+        throw new Error(`Definition failed for '${name}': ` +
+            `The constructor is already used.`);
+      }
+
+      // 8:
+      var localName = name;
+
+      // 9, 10: We do not support extends currently.
```
sorvell: Is type extension planned? If so, maybe we can add a note somewhere that it's a planned feature?

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+                definition.connectedCallback.call(node);
+              }
+            }
+          }
+          if (node.shadowRoot) {
+            // TODO(justinfagnani): do we need to check that the shadowRoot
+            // is observed?
+            this._addNodes(node.shadowRoot.childNodes);
+          }
+          if (node.tagName === 'LINK') {
```
kevinpschaaf: This HTML import walk doesn't consider link de-duplication; as such, it performs an _excessive_ amount of walking through duplicated links in the dependency tree (a basic smoke test that would otherwise load 31 imports had them walked 2000+ times).  Could be mitigated by hanging a "has been walked" flag, or if needed a "last walked" monotonically increasing id off of `link.import`, and skipping links that have already been walked for this mutation, for example.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+  ];
+
+  for (var i = 0; i < htmlElementSubclasses.length; i++) {
+    var ctor = window['HTML' + htmlElementSubclasses[i] + 'Element'];
+    if (ctor) {
+      ctor.prototype.__proto__ = win.HTMLElement.prototype;
+    }
+  }
+
+  // patch doc.createElement
```
kevinpschaaf: It's critical to also patch `document.importNode` to ensure synchronous upgrade of stamped templates.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+    /**
+     * @private
+     */
+    _handleMutations: function(mutations) {
+      for (var i = 0; i < mutations.length; i++) {
+        var mutation = mutations[i];
+        if (mutation.type === 'childList') {
+          // Note: we can't get an ordering between additions and removals, and
+          // so might diverge from spec reaction ordering
+          this._addNodes(mutation.addedNodes);
```
kevinpschaaf: As implemented, `_addNodes` always connects elements after customization, which is not correct for HTML imports, or additions of shadow root children of detached hosts.

You can bind `root` into `_handleMutations`, and then tell `_addNodes` to only connect if the root is `window.document`, or if `root.host` is currently attached.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+
+        var walker = createTreeWalker(root);
+        do {
+          var node = /** @type {HTMLElement} */ (walker.currentNode);
+          var definition = this._definitions.get(node.localName);
+          if (definition) {
+            if (!node.__upgraded) {
+              this._upgradeElement(node, definition, true);
+            }
+            if (node.__upgraded && !node.__attached) {
```
sorvell: Is the node guaranteed to be in the document at this point? If not, then should not call `connectedCallback`. For example, this should (probably) not be called in HTMLImports.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+    _setNewInstance: function(instance) {
+      this._newInstance = instance;
+    },
+
+    /**
+     * Observes a DOM root for mutations that trigger upgrades and reactions.
+     * @private
+     */
+    _observeRoot: function(root) {
+      root.__observer = new MutationObserver(this._handleMutations.bind(this));
```
sorvell: Bail if the root is already observed?

src/CustomElements/v1/CustomElements.js
sorvell: +1 need to only handle `rel="import"` `link` elements. If you get a `load` even, I believe `import` is guaranteed. Can probably use `.children` here.

The `.import` property may point to an already observed document due to HTMLImports document de-duping. Must avoid re-observing in this case!

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+
+        if (!isElement(root)) {
+          continue;
+        }
+
+        // Since we're detatching this element from an observed root, we need to
+        // reobserve it.
+        // TODO(justinfagnani): can we do this in a microtask so we don't thrash
+        // on creating and destroying MutationObservers on batch DOM mutations?
+        this._observeRoot(root);
```
sorvell: Dubious that this is worth it. This is for `innerHTML`? Consider not supporting this or making it manual and/or consider patching strategy outlined in other comment.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+          this._removeNodes(mutation.removedNodes);
+        }
+      }
+    },
+
+    /**
+     * @param {NodeList} nodeList
+     * @private
+     */
+    _addNodes: function(nodeList) {
```
sorvell: Consider exposing a version of this as a manual backstop for upgrading elements that the polyfill misses.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+      if (callConstructor) {
+        this._setNewInstance(element);
+        element.__upgraded = true;
+        new (definition.constructor)();
+        console.assert(this._newInstance == null);
+      }
+
+      var observedAttributes = definition.observedAttributes;
+      if (definition.attributeChangedCallback && observedAttributes.length > 0) {
+        this._attributeObserver.observe(element, {
```
sorvell: If we go with the patching strategy, then overriding `setAttribute`/`removeAttribute` is likely good enough.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+
+  var origHTMLElement = win.HTMLElement;
+  win.HTMLElement = function HTMLElement() {
+    var customElements = win['customElements'];
+    if (customElements._newInstance) {
+      var i = customElements._newInstance;
+      customElements._newInstance = null;
+      return i;
+    }
+    if (this.constructor) {
```
sorvell: Not sure I grok this case. Can you explain?

sorvell: Suggest adding a comment here indicating that `_newInstance` is for upgrade and the other path is for creation.

src/CustomElements/v1/CustomElements.js
```diff
@@ -0,0 +1,597 @@
...
+
+  var rawCreateElement = doc.createElement;
+  doc._createElement = function(tagName, callConstructor) {
+    var customElements = win['customElements'];
+    var element = rawCreateElement.call(doc, tagName);
+    var definition = customElements._definitions.get(tagName.toLowerCase());
+    if (definition) {
+      customElements._upgradeElement(element, definition, callConstructor);
+    }
+    customElements._observeRoot(element);
```
sorvell: Still dubious about this...

src/CustomElements/v1/native-shim.js
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * This shim allows elements written in, or compiled to, ES5 to work on native
```
sorvell: Is the alternative to this to require transpiling for ES5 browsers that use the polyfill?

If we're going to bother with this shim, it'd be great to make it transparent to use.

gulpfile.js
justinfagnani: Closure gives us much better optimizations than other build tools, and gives us ES6->ES5 compilation, which uglify doesn't.

src/CustomElements/v1/README.md
addyosmani: After re-reading this a few times I'm similarly unclear on how the shim is intended to be used. I'm unsure if the final build of CE V1 will support the shim by default or if I need to include a separate file to get it working in both environments.

src/CustomElements/v1/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+`new.target` isn't even feature detectable, since it's a syntax error in ES5. Because of this, the polyfill can't check `new.target` first and fallback to `this.constructor`. This also means that ES5-style constructors can't conditionally make a "super" call to the HTMLElement constructor (with `Reflect.construct`) in non-ES6 environments to be compatible with native Custom Elements.
+
+To allow for elements that work in both ES5 and ES6 environments, we provide a shim that overrides the HTMLElement constructor and calls `Reflect.construct` with either the value of `new.target` or `this.constructor`. This shim can only be executed in ES6 environments.
+
+## Building
+
+The Custom Elements V1 polyfill does not use the same module or build system as
+the other webcomponentsjs polyfills, and its build is not integrated into the default build tasks yet.
+
```
addyosmani: Is the intention to integrate it back in with the rest of the build system shortly after this initial PR lands?

src/CustomElements/v1/CustomElements.js
addyosmani: An escape hatch for forcing the polyfill if absolutely needed would be super useful. +1.

gulpfile.js
addyosmani: Justin is correct that Closure does still provide a few optimizations other tools don't (see https://github.com/samccone/The-cost-of-transpiling-es2015-in-2016 for one comparison). 

I am concerned about having a separate build process and pipeline to the rest of the project potentially confusing contributors. I'd love to see us either rally around closure as other parts of the repo get updated (e.g SD V1) or better integrating this work with the existing build system.

trusktr: Using ES6 modules (or CommonJS) would encourage much more contribution. Developers could import the modules into any runtime or build environment that supports ES6 Modules, and won't be strictly required to use your build system.

Currently, developers are required to use your build flow or make a brand new one in order to create something usable, and it is difficult because they have to make sure to load the files in the correct order. Using a module system solves that problem.

If you switch to a module system, you can still make a global by creating an index file that imports whatever it needs then setting those things onto `window`.

I would highly recommend making the switch.

src/CustomElements/v1/native-shim.js
justinfagnani: No, this shim is for elements that _are_ transpiled to ES5. That is, transpiling is not an alternative to the shim, but causes the shim to be necessary.

src/CustomElements/v1/README.md
justinfagnani: The existing build system is very custom, and given that this version is one file, I don't see the need for it. I can integrate it further with the gulp file so that it build by default.

src/CustomElements/v1/README.md
justinfagnani: The shim is needed in for transpiled elements in browsers that support CustomElements v1 natively. It'll only be able to be loaded in browsers that support classes and `new.target`, so it'll require some browser sniffing (feature detection is basically impossible for `new.target`).

If an app controls all of its elements and dependencies, and they're all written in ES6 classes, then instead of using the shim it can use the ES6 version of the elements, and only send the ES5 version to older browsers. This may end up being the case all of the time, since CEv1 strongly suggests writing elements with classes. If so, the shim will be rarely or never used.

justinfagnani: Oh, and the shim can never be included by default with the polyfill. `new.target` is a syntax error if it's not supported, so you can't feature detect it. If it were included in the polyfill, the whole polyfill would fail to load.

wessberg:
This:
```javascript
const testComponent = new TestComponent();
document.body.appendChild(testComponent);
```

Calls the constructor twice. This is an issue because everything that happens in the constructor will happen twice.

The implementation does take this scenario into account between lines 464 - 467:
```javascript
if (this.constructor) {
      var tagName = customElements._constructors.get(this.constructor);
      return doc._createElement(tagName, false);
    }
```

However, the `_addNodes` method matches the node and calls `_upgradeElement` with a truthy `callConstructor` value.

A temporary hotfix of mine:
1) Add a `_calledConstructors` WeakSet in the `CustomElementsRegistry` method.
2) Check if the constructor has been called before in `upgradeElement`, add the element to the set if it hasn't and call the constructor:
```javascript
const hasCalledConstructor = this._calledConstructors.has(element);
	if (!hasCalledConstructor) this._calledConstructors.add(element);
	if (callConstructor) {
		this._setNewInstance(element);
		element.__upgraded = true;
		if (!hasCalledConstructor) {
			new definition.constructor;
		}
		this._newInstance = null;
	}
...
```

src/CustomElements/v1/CustomElements.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: I'd rather wait till this comes up. The polyfill shouldn't miss any elements.

src/CustomElements/v1/CustomElements.js
justinfagnani: Added an assert, since this should only be called if the root isn't already observed.

src/CustomElements/v1/CustomElements.js
justinfagnani: Done, and added tests

src/CustomElements/v1/CustomElements.js
justinfagnani: Any reason? Seems like they should be documented on the class.

src/CustomElements/v1/README.md
justinfagnani: done

src/CustomElements/v1/README.md
justinfagnani: Added some clarifying text, I hope.

src/CustomElements/v1/README.md
justinfagnani: This whole file is for people interested in the implementation, otherwise we need to update the repo's README.

gulpfile.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: it's already null in that case

justinfagnani: it's already null in that case

src/CustomElements/v1/CustomElements.js
justinfagnani: added much code and tests for imports

gulpfile.js
justinfagnani: @trusktr This library neither imports or exports anything in the module sense. Because it's a polyfill, it's directly patching prototypes and defining a global. I'm not sure modules would buy us anything, while they would still require compilation.

I'm not sure why anyone would need to define their own workflow. We will distribute the built files.

justinfagnani: @addyosmani Since I hadn't seen or worked with the rest of the polyfills, I decided during development to just focus on a from scratch and simpler (to me anyway) build workflow that gets us maximum optimizations. As we get closer to releasing this, I'm fully in favor of integrating better.

I still haven't fully ingested the custom build system used in the rest of the polyfills. I'd personally be more inclined to move everything to Closure and not have to maintain something custom. I'm not sure how much work that is at the moment.

gulpfile.js
justinfagnani: Build integration with the rest of the polyfills, TBD (see the above thread with @addyosmani).

I would actually like to see the polyfills separated. To me it's clearer, and given a mix of environments with different needs, I think it's easier to know what to include. To use Polymer on Safari, for instance, we'll only need Custom Elements and HTML Imports. On older versions of Chrome, just Custom Elements.

src/CustomElements/v1/CustomElements.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: I'm going to address all the other issues before coming back around on observation vs patching. I'd like to see if we can separate the registry, upgrade and reaction parts from the observation parts and give some choice here.

src/CustomElements/v1/CustomElements.js
justinfagnani: done

src/CustomElements/v1/CustomElements.js
justinfagnani: done

gulpfile.js
trusktr: > We will distribute the built files.

You can still do that even if it's written with ES6 modules. I don't remember why, but I tried to get this running much earlier on when you started the work on this, but I couldn't get it working and I figured it was because you didn't hook it into your build system, and I didn't want to spend the time to figure it out. It may not be the case anymore if this file runs completely standalone.

If I can import this file directly into my ES6 environment, then I'll be happy (because that is the way in which JavaScript is recommended to be used now, while global-style code is discouraged though not disallowed).

In fact, I believe now that ES6 modules are official, it'd be nice if all new APIs were `import`able and never global. In this case, `customElements` would not be global, it would be imported. `import customElements from 'window/CustomElements'` or similar. This would be awesome and is the direction everything should go in.

src/CustomElements/v1/CustomElements.js
rictic: Mainly because `@property` isn't recognized by closure compiler: https://groups.google.com/forum/#!topic/closure-compiler-discuss/vZNIh9vkJPU

It instead looks for either `@type`, `@public`, `@private`, or `@protected` in `this.foo` statements or assignments in the constructor.

addyosmani:
Should this PR be closed now that we've got https://github.com/webcomponents/custom-elements? (imagine we might want to add that as a dependency to this repos package.json once we're ready)

tests/ce-upgrade-order.html
```diff
@@ -11,17 +11,23 @@
...
-      chai.assert.equal(span.textContent, 'Róbert Viðar Bjarnason');
+    test('upgrade order', function() {
+      var order = [];
+      class XFoo extends HTMLElement {
+        connectedCallback() {
+          order.push(this.id);
+        }
+      }
+      window.customElements.define('x-foo', XFoo);
+      assert.deepEqual(order, ['import', 'main'], 'elements are upgraded in imports before main document');
```
valdrinkoshi: This should wait for `HTMLImports.whenReady` in order to work, right?

azakus:
IT IS HAPPENING

notwaldorf:
👀😳😎

zanona:
Congrats and thanks for the great work guys! You are awesome!

PR #522, opened 2016-04-13
Make sure IE11 importNode workaround for doc fragments is patch on importNode

### LINKED ISSUE #518: IE11 Unable to get property 'replaceChild' of undefined or null reference
v0.7.21 causes the following exception to be thrown in IE11:

```js
SCRIPT5007: Unable to get property 'replaceChild' of undefined or null reference
File: webcomponents-lite.js, Line: 1037, Column: 9
```

The debugger says that `t` is `[object HTMLUnknownElement]`. I've been able to determine which custom element it's supposed to be based on `classList`. This seems to happen only for custom elements which extend native ones (but does not happen for all custom elements extending native ones). Ex `my-bar` and `my-foo` extend `a`, but only `my-bar` triggers this exception (`my-bar` always does and `my-foo` never does). Possibly a red-herring, but alphabetical order seems to matter (possibly due to our build process, which runs in alphabetical order, so my-bar would be included before my-foo).

v0.7.20 causes some Hierarchy error. v0.7.19 appears to be the last stable version.

Unfortunately, IE crashes when I try to create a jsbin for this.

azakus:
Fixes #518

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
-sudo: false
+sudo: required
+dist: trusty
 node_js: 4
 addons:
   sauce_connect: true
@@ -7,10 +8,8 @@ addons:
   apt:
     sources:
     - google-chrome
-    - ubuntu-toolchain-r-test
     packages:
     - google-chrome-stable
-    - g++-4.8
 before_script:
 - export PATH=$PWD/node_modules/.bin:$PATH
 script:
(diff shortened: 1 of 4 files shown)


azakus:
Hmm, Safari 7 tests seem busted. Need to investigate.

ebidel:
Is anyone taking a look at this? We'd like to get back on the latest wc.js version and test.

azakus:
Latest patch fixes Edge 13 tests as well, need to file a bug on them, as `template.cloneNode(true)` is totally busted.

kevinpschaaf:
LGTM.

PR #525, opened 2016-04-16
Keep prefix attribute selectors while shimming

### LINKED ISSUE #207: [ShadowCSS] Wrong handling of prefix attributes ([attr^=value])
If a style contains an `[attr^=value]` selector, the "^" will be removed in the output.

This happens because `scopeCssText()` calls `convertShadowDOMSelectors()` which replaces all `shadowDOMSelectorsRe` with a space. `/\^/g` is the 2nd regexp in this array.

Side question: what are the `^` and `^^` selectors ?

AMoo-Miki:
ShadowCSS converts all prefix attribute selectors `[attr^=value]` to exact selectors `[attr=value]`.

The bug was caused due to a bad regular expression which intended to remove `^`. To maintain status quo but fix the issue, I have simply put in a negative look-ahead for `=`.

Fixes #207


azakus:
LGTM

PR #537, opened 2016-05-25
Use function expression assignment over declaration in blocks

rictic:
Function declarations inside blocks don't behave like it looks like they do, and this isn't actually legal ES5 strict, so it can cause problems downstream after minification.


TimvdLippe:
Appears to be fixed in https://github.com/webcomponents/html-imports/blob/master/src/html-imports.js

PR #538, opened 2016-05-25
Added alternative workaround for getting wrapped `document` in a generic way

nazar-pc:


nazar-pc:
Can anyone take a look at this trivial addition to documentation?

TimvdLippe:
This method has been removed, so I think this addition is no longer necessary.

PR #544, opened 2016-05-31
V1 npm

trusktr:
This works towards supporting the repo itself as an NPM package, so that it's possible to put something like

```json
"webcomponentsjs": "webcomponents/webcomponentsjs#v1"
```

inside of a dependent's package.json `dependencies`.

This is needed so we can import the `v1` branch from GitHub into a project by doing

```bash
npm install
```
```js
import `webcomponents.js`
```

instead of

```bash
npm install
cd node_modules/webcomponents.js
gulp build
```
```js
import `webcomponents.js/dist/webcomponents.js`
```

### DIFF
--- a/externs/html5.js
+++ b/externs/html5.js
@@ -1,0 +1,1 @@
+MutationObserver.prototype.takeRecords = function() {};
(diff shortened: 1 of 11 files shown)


trusktr:
Oops, wrong branch

PR #545, opened 2016-05-31
NPM support.

trusktr:
This works towards supporting the repo itself as an NPM package, so that it's possible to put something like

``` json
"webcomponents.js": "webcomponents/webcomponentsjs#v1"
```

inside of a dependent's package.json `dependencies`.

This is needed so we can import the `v1` branch from GitHub into a project by doing

``` bash
npm install
```

``` js
import `webcomponents.js`
```

instead of

``` bash
npm install
cd node_modules/webcomponents.js
gulp build
```

``` js
import `webcomponents.js/dist/webcomponents.js`
```

I tested this by renaming the package to random `test-aksdfhasdfkjqhwecuasdfkashjf` and publishing: https://www.npmjs.com/package/test-aksdfhasdfkjqhwecuasdfkashjf

This means that running `npm publish` will automatically build before publishing, and will only publish the `dist` folder, and `require('webcomponents.js')` will load the dist file.

This is also useful for `npm link` workflows, as that will also run `npm install && npm prepublish` behind the scenes in order to make local development easier.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,6 @@
+/*
+!/dist
+!/CONTRIBUTING.md
+!/LICENSE.md
+!/README.md
+!/package.json
--- a/externs/html5.js
+++ b/externs/html5.js
@@ -1,0 +1,1 @@
+MutationObserver.prototype.takeRecords = function() {};
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,17 +12,19 @@
 
 'use strict';
 
-var
-  audit = require('gulp-audit'),
-  concat = require('gulp-concat'),
-  exec = require('child_process').exec,
-  fs = require('fs'),
-  gulp = require('gulp'),
-  header = require('gulp-header'),
-  path = require('path'),
-  runseq = require('run-sequence'),
-  uglify = require('gulp-uglify')
-;
+var audit = require('gulp-audit');
+var compilerPackage = require('google-closure-compiler');
+var concat = require('gulp-concat');
+var exec = require('child_process').exec;
+var fs = require('fs');
+var gulp = require('gulp');
+var header = require('gulp-header');
+var path = require('path');
+var rename = require('gulp-rename');
+var runseq = require('run-sequence');
+var uglify = require('gulp-uglify');
+
+var closureCompiler = compilerPackage.gulp();
 
 // init tests with gulp
 require('web-component-tester').gulp.init(gulp);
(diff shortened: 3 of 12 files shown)


trusktr:
I tested this by renaming the package to random `test-aksdfhasdfkjqhwecuasdfkashjf` and publishing: https://www.npmjs.com/package/test-aksdfhasdfkjqhwecuasdfkashjf

trusktr:
This means that running `npm publish` will automatically build before publishing, and will only publish the `dist` folder, and `require('webcomponents.js')` will load the dist file.

trusktr:
This is also useful for npm link workflow, as that will also run npm install && npm prepublish behind the scenes in order to make local development easier.

trusktr:
I haven't checked how this affects the Bower workflow though, but I'd encourage end users to use NPM + Webpack/Browserify/etc and transition away from Bower.

trusktr:
Oops, after installing a second time, the dist folder doesn't appear to be there. Let me see...

trusktr:
Oh, because installing from GitHub doesn't run the `prepublish` script. Hmmm...

trusktr:
Well, apparently ["patches welcome"](https://github.com/npm/npm/issues/3055) on NPM to make that work. So, what I'll do for now is just publish to a new NPM repo temporarily, `@trusktr/webcomponents.js`.

trusktr:
No luck yet. Will post back when it works.

justinfagnani:
We're already published on npm at: https://www.npmjs.com/package/webcomponents.js

justinfagnani:
I read too fast and skipping the github dependency part. It seems very common to build to a dist folder and publish from there. I assume that this always breaks github dependencies, and so people check out the repo and `npm link`. Is that a usable workaround for you?

trusktr:
That workaround is a bit more involved. It'd be nice to simply list the `repo#commit` in package.json.

I'll get back to testing this change, I wasn't able to get `customElements.define` to work for some reason, not sure if it's because I'm importing the v1 files directly (not the dist files, since `gulp build` doesn't build the v1 files yet, so that change can also be made in this PR).

Or, maybe I'm using the API wrong? I previously had something like this with v0:

```js
class MyElement extends HTMLElement {
  // ..
}

export default
document.registerElement('my-element', MyElement)
```

which I converted to the following which caused the element to stop being upgraded (as if I never registered it):

```js
export default
class MyElement extends HTMLElement {
  // ..
}

customElements.define('my-element', MyElement)
```

If I understand v1, the major difference is that the original constructor can be used unlike with v0 where the constructor needs to be received from the `registerElement` call, which is why in the second example I am exporting the original class directly rather than taking a value from the `define` call (which is undefined).

Am I doing it correctly?

trusktr:
Oh, and I'm importing v1 like this:

```js
import 'webcomponents.js-v1/src/CustomElements/v1/native-shim'
import 'webcomponents.js-v1/src/CustomElements/v1/CustomElements'
```

where `webcomponents.js-v1` is the package on NPM that I'm using to test these changes. I was going to import from `dist`, but then I realized `gulp build` hadn't handled the v1 files yet.

TimvdLippe:
The polyfills are now published on NPM, so this issue should be resolved now.

PR #553, opened 2016-06-07
Add support for old Opera/Presto engine.

kevinpschaaf:
Adds minimal support for old Opera Presto engine:
- need to use `oMatchesSelector`
- need to use "fakeLoad" logic for `<style>` elements added for IE
- need to fix up URL's in linked imports, as XHR'ed docs do not respect the baseURI

### DIFF
--- a/src/HTMLImports/parser.js
+++ b/src/HTMLImports/parser.js
@@ -14,6 +14,7 @@ var path = scope.path;
 var rootDocument = scope.rootDocument;
 var flags = scope.flags;
 var isIE = scope.isIE;
+var isOldOpera = scope.isOldOpera;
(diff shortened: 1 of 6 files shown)


src/HTMLImports/parser.js
```diff
@@ -191,7 +192,7 @@ var importParser = {
 
     // NOTE: IE does not fire "load" event for styles that have already loaded
```
justinfagnani: update comment

justinfagnani:
LGTM

addyosmani:
LGTM2. @justinfagnani was there more work that needed to be done here?

TimvdLippe:
Closing due to age. If you still want to incorporate these changes, they have to be changed in https://github.com/webcomponents/html-imports/blob/master/src/html-imports.js

Issue #556, opened 2016-06-08
TreeWalker fixes for IE


treshugart:
IE 11 requires the last argument to [`document.createTreeWalker()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker) even though it's being removed. If it's not provided we get an error stating "Argument not optional".

In fixing that, you must also provide the third argument. In IE 11 you can pass `null`, however, I believe this fails in IE9 and 10 because they require a you specify a function for `acceptNode`. Are we supporting these (assuming not due to the usage of classes)? If so it requires something like:

```js
function createTreeWalker (root) {
  // Accept all currently filtered elements.
  function acceptNode () {
    return NodeFilter.FILTER_ACCEPT;
  }

  // Work around Internet Explorer wanting a function instead of an object.
  // IE also *requires* this argument where other browsers don't.
  const safeFilter = acceptNode;
  safeFilter.acceptNode = acceptNode;

  return doc.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, safeFilter, false);
}
```

Instead of just:

```js
function createTreeWalker (root) {
  return doc.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, null, false);
}
```

Once you provide IE11 with the correct arguments, it complains with "Unexpected call to method or property access." when calling `walker.nextNode()`. The fix for this is to only walk trees for nodes that are `Node.ELEMENT_NODE`.

treshugart:
This was fixed in #560.

lukeed:
Just ran into this at work. Not sure _how_ you found this out, but thank you @treshugart 😆

PR #560, opened 2016-06-08
#556 - fix tree walker in IE11

treshugart:


src/CustomElements/v1/CustomElements.js
```diff
@@ -61,6 +61,10 @@ var CustomElementDefinition;
     return reservedTagList.indexOf(name) !== -1;
   }
 
+  function createTreeWalker (root) {
```
justinfagnani: style: remove space after function name

src/CustomElements/v1/CustomElements.js
```diff
@@ -61,6 +61,10 @@ var CustomElementDefinition;
...
+    return doc.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, null, false);
```
justinfagnani: Previously I think you had a comment about how IE needed these arguments. That was useful.

justinfagnani:
LGTM with two very small nits

Issue #557, opened 2016-06-08
Unnecessary rawCreateElement.call()


treshugart:
It's already bound to the document, so we can just call it directly: `rawCreateElement(name)`.

treshugart:
Fixed in #561.

PR #561, opened 2016-06-08
#557 - call rawCreateElement() directly

treshugart:


src/CustomElements/v1/CustomElements.js
```diff
@@ -296,7 +296,7 @@ var CustomElementDefinition;
   var rawCreateElement = doc.createElement.bind(document);
   doc._createElement = function(tagName, callConstructor) {
     var customElements = win['customElements'];
-    var element = rawCreateElement.call(document, tagName);
+    var element = rawCreateElement(tagName);
```
justinfagnani: Actually, I've seen claims of benchmarks from former V8 engineers that `.call` is still an order of magnitude faster than a bound function, even after the very latest bound function optimizations. Maybe we should remove the `.bind()` call above.

justinfagnani:
@treshugart oops, I didn't see that you pushed a new commit!

PR #571, opened 2016-06-18
Remove not required else block

mitogh:
Simplify a block by removing not required else, since the line is always executed on both blocks (if /else) there is no need to define such blocks.

### DIFF
--- a/src/WebComponents/build/boot-lite.js
+++ b/src/WebComponents/build/boot-lite.js
@@ -37,14 +37,12 @@
       }
     }
     // log flags
+    flags.log = {};
     if (flags.log && flags.log.split) {
       var parts = flags.log.split(',');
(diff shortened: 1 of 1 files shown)


src/WebComponents/build/boot-lite.js
```diff
@@ -37,14 +37,12 @@
       }
     }
     // log flags
+    flags.log = {};
```
justinfagnani: This isn't a behavior preserving change. With this change, `flags.log` will always be defined and `flags.log.split` will always be undefined, so the if body will never be executed.

mitogh: You're right, sorry for the inconvenience and thank you for taking the time for review this.

I'm going to close this PR if you don't mind.

Thanks.

PR #583, opened 2016-07-28
[v1] fix error in handling of HTML imports

bash:


justinfagnani:
This should be fixed along with a lot of other link/import handling in the `v1` branch now. Thanks for the PR.

Issue #112, opened 2014-12-09
new querySelector logic broken if no <content> element


tjsavage:
<a href="https://github.com/jakemac53"><img src="https://avatars.githubusercontent.com/u/984921?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jakemac53](https://github.com/jakemac53)**
_Tuesday Aug 26, 2014 at 20:55 GMT_
_Originally opened as https://github.com/Polymer/ShadowDOM/issues/495_

---

https://github.com/Polymer/ShadowDOM/commit/14bc904433b71dd8c6de0091c2fa4fccdb66f960 skips some tests which are exposing real errors. Specifically, if you do not have a content element in  your polymer elements template, but somebody does supply content for your element, then they can no longer query for those elements (see the querySelectorAll test). While these elements will not be rendered they should still be queryable.

nazar-pc:
I did partial fix for this issue in #624 (and unlocked some previously skipped tests).
My guess is that no one will ever work on this for v0, so PR in its current form actually makes sense.

PR #589, opened 2016-08-06
Range shadow fix

nazar-pc:
Fix for Range polyfill in cases when `range.startContainer` and/or `range.endContainer` are projected into Shadow Root, but are themselves inside of another Shadow Root.

Test cases are also slightly adjusted to produce more useful output on failure, wich I found useful during debugging.

@addyosmani, @eeid26 and @jmesserly for review, since you were active reviewers on related PR #264

### DIFF
--- a/src/ShadowDOM/wrappers/Range.js
+++ b/src/ShadowDOM/wrappers/Range.js
@@ -24,6 +24,13 @@
   var ShadowRoot = scope.wrappers.ShadowRoot;
 
   function getHost(node) {
+    var parent = node;
+    while (parent = parent.parentNode) {
+      if (parent.shadowRoot) {
+        return null;
+      }
+    }
+
     var root = getTreeScope(node).root;
     if (root instanceof ShadowRoot) {
       return root.host;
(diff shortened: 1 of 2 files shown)


nazar-pc:
Ping here, can anyone take a look at these important Range fixes?

src/ShadowDOM/wrappers/Range.js
```diff
@@ -24,6 +24,13 @@
   var ShadowRoot = scope.wrappers.ShadowRoot;
 
   function getHost(node) {
+    var parent = node;
+    while (parent = parent.parentNode) {
+      if (parent.shadowRoot) {
```
justinfagnani: Can you explain and document why the host should be `null` if any ancestor has a `shadowRoot`?

nazar-pc: This is for consistency with code at the end of the function.
Function is expected to get Host, but if there is no suitable host, then `null` is returned (which below in `getHost(node) || node` effectively results in node itself).

nazar-pc:
I've faced #112 in my project, which seems to be related to https://github.com/Polymer/polymer/issues/3809 and apparently have the same root as this issue.

Will work on #112 today (I really need to have it fixed), hopefully wouldn't take more than 10 hours in debugger:)

nazar-pc:
Any objections on this PR? It works great for me for quite some time now, big TinyMCE works fine with this fix, but some features doesn't without this fix.

TimvdLippe:
Closing this PR due to age. If you would still like to propose a fix, please do that at https://github.com/webcomponents/shadydom (but I think this has been fixed already in the latest version of the polyfills)

PR #591, opened 2016-08-11
Fix for removeEventListener bug

jbwyme:
`target.addEventListener_(type, dispatchOriginalEvent, true)` is used to bind listeners to the target regardless if `useCapture` is true. That's ok until you go to remove:

```
if (found && count === 1) {
    var target = getTargetToListenAt(this);
    target.removeEventListener_(type, dispatchOriginalEvent, true);
}
```

This means if you bind an event listener using `useCapture: true`, bind another event listener using `useCapture: false`, and then remove the second listener it will stop listening to the target element even though we still expect to have the first event listener bound.

### DIFF
--- a/src/ShadowDOM/wrappers/events.js
+++ b/src/ShadowDOM/wrappers/events.js
@@ -777,7 +777,7 @@
       listeners.push(listener);
 
       var target = getTargetToListenAt(this);
-      target.addEventListener_(type, dispatchOriginalEvent, true);
+      target.addEventListener_(type, dispatchOriginalEvent, capture);
     },
     removeEventListener: function(type, fun, capture) {
       capture = Boolean(capture);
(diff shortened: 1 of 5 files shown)


src/ShadowDOM/wrappers/events.js
```diff
@@ -797,7 +797,7 @@
 
       if (found && count === 1) {
         var target = getTargetToListenAt(this);
-        target.removeEventListener_(type, dispatchOriginalEvent, true);
+        target.removeEventListener_(type, dispatchOriginalEvent, capture);
```
jbwyme: An alternative solution to this: If we need `useCapture` to always be `true`, we could rework the logic here so that we get the count of ALL bound listeners instead of `listeners[i].capture === capture`

jbwyme:
@justinfagnani any chance you could take a look at this?

justinfagnani:
I'm not sure why the original code would have hardcoded useCapture: `true`. That seems odd enough that there may have been a reason. @sorvell any idea?

TimvdLippe:
Appears to be fixed in https://github.com/webcomponents/shadydom/blob/391af315e508e03fd5224eb8b4f6fdb76a1a86b5/src/patch-events.js#L306-L409

PR #594, opened 2016-08-14
Simulate native `constructor` descriptor

WebReflection:
This change uses a patched `constructor` descriptor that is similar to a native one and fixes https://github.com/skatejs/web-components/issues/2

### DIFF
--- a/src/CustomElements/v1/CustomElements.js
+++ b/src/CustomElements/v1/CustomElements.js
@@ -489,8 +489,10 @@ var CustomElementDefinition;
       }
       throw new Error('unknown constructor. Did you call customElements.define()?');
     }
-    patched.prototype = Object.create(orig.prototype);
-    Object.defineProperty(patched.prototype, 'constructor', {value: patched});
+    patched.prototype = Object.create(
+      orig.prototype,
+      {constructor: {configurable: true, value:patched, writable: true}}
+    );
   }
 
   patchConstructor('HTMLElement');


WebReflection:
P.S. the Travis failure seems to be completely unrelated `¯\_(ツ)_/¯`

bradleyayers:
LGTM

src/CustomElements/v1/CustomElements.js
```diff
@@ -489,8 +489,10 @@ var CustomElementDefinition;
       }
       throw new Error('unknown constructor. Did you call customElements.define()?');
     }
-    patched.prototype = Object.create(orig.prototype);
-    Object.defineProperty(patched.prototype, 'constructor', {value: patched});
+    patched.prototype = Object.create(
+      orig.prototype,
+      {constructor: {configurable: true, value:patched, writable: true}}
```
treshugart: I believe `constructor` should also be `enumerable: false`.

WebReflection: it is indeed, that's what the default is for every property as defined in specs.

treshugart: True; my bad.

bradleyayers:
What's left to do with this change before you're comfortable merging it?

WebReflection:
if you're asking to me :P

  1. I cannot merge
  2. I have no idea why Travis is failing

Otherwise I would've merged right away ;-)

treshugart:
We should probably write a test ensuring the constructors we patch are `configurable`, `writable` and non-`enumerable`. I'll try and check why the build is failing; it looks like most current red builds have the same issue.

WebReflection:
@treshugart to be honest, it seems like you want to test that `Object.create` second argument works as expected, since that's the only operation that change but that's also *not* your unit-test responsibility.

One shouldn't test native/third parts operations, it's like saying "we should write a test that validate that 1 + 1 === 2" ;-)

treshugart:
Disagree. If code can break something it should be tested.

WebReflection:
that's my point. This code cannot break anything at all, I've added two properties to a descriptor handled by native.

Anyway, feel free to test `Object.create`

WebReflection:
P.S. if you are talking about configuration expectations, then I agree, but that should've been written already in the first place ;-)

treshugart:
Not critiquing your code and stoked to have a fix, it would just be nice to
have a corresponding test. I'll see if I can commit one this weekend.
Thanks again :)

justinfagnani:
Thanks for the PR. I copied this change over to the `v1` branch.

Issue #595, opened 2016-08-14
[v1] consistent constructor


WebReflection:
It seems like it's possible to `class MyElement extends HTMLElement {}` and then `document.body.appendChild(new MyElement)` after defining the element through `customElements.define('my-element', MyElement)`.

However, if I try to do the same, using `{extends: ...}` or not, I cannot create new custom elements in the same way.

``` js
class MyAnchor extends HTMLAnchorElement {}
customElements.define('my-a', MyAnchor);
// or
customElements.define('my-a', MyAnchor, {extends: 'a'});

// will fail
document.body.appendChild(new MyAnchor);
```

I wonder if it's in the pipeline to monkey-patch all HTML constructors like it's done already only for few of them, or if the usage of `new CustomElement` instead of `document.createElement('a', {is: 'my-a'})` will be discourage due polyfill limits (I personally wouldn't polyfill every DOM methods that might accept a generic node as one of its arguments).

Thanks for any sort of clarification.

treshugart:
I think the issue you're experiencing is that `extends` hasn't been implemented yet: https://github.com/webcomponents/webcomponentsjs/blob/v1/src/CustomElements/v1/CustomElements.js#L143

Subclassing `HTMLAnchorElement` *should* still work, though: https://github.com/webcomponents/webcomponentsjs/blob/v1/src/CustomElements/v1/CustomElements.js#L477

I know there's currently still unresolved contention about `is` and `extends` as WebKit is pretty hard-fast about not implementing it, so I'm not sure what the polyfill will work to support. I assume it's on the roadmap as it's still in the custom element spec.

WebReflection:
I've probably badly explained the issue. Beside `extends` and `is` definition variant, the following code will fail (and every other code that has [not been patched like HTMLElement in here](https://github.com/webcomponents/webcomponentsjs/blob/v1/src/CustomElements/v1/CustomElements.js#L470))

```js
// create the class
class MyA extends HTMLAnchorElement {}

// define its name
customElements.define('my-a', MyA);

// and fail ...
document.body.appendChild(new MyA);
```
If I do the same with the patched `HTMLElement` the result is the expected one.
```js
class MyElement extends HTMLElement {}
customElements.define('my-el', MyElement);
document.body.appendChild(new MyElement); // OK
```
[...]

PR #596, opened 2016-08-15
#595 - feat: add initial support for customised built-in elements.

treshugart:
Implements initial support for #595.

I'd like to write more tests, but this is what I could come up with for tonight.

### DIFF
--- a/src/CustomElements/v1/CustomElements.js
+++ b/src/CustomElements/v1/CustomElements.js
@@ -550,11 +550,9 @@ var CustomElementDefinition;
     const name = 'HTML' + htmlElementSubclasses[i] + 'Element';
     const ctor = window[name];
     if (ctor) {
-      win[name] = class extends HTMLElement {
-        constructor() {
-          super()
-        }
-      };
+      win[name] = class extends HTMLElement {};
+
+      // Required so that document.createElement() instanceof checks pass.
       ctor.prototype.__proto__ = HTMLElement.prototype;
       ctor.prototype.__proto__ = win[name].prototype;
     }


src/CustomElements/v1/CustomElements.js
```diff
@@ -103,7 +103,7 @@ var CustomElementDefinition;
 
     // HTML spec part 4.13.4
     // https://html.spec.whatwg.org/multipage/scripting.html#dom-customelementsregistry-define
-    define: function(name, constructor, options) {
+    define: function(name, constructor, options = {}) {
```
treshugart: I think this is probably `null` in the spec, it just makes using it simpler. Not attached to keeping the default value if spec consistency here is super important.

src/CustomElements/v1/CustomElements.js
```diff
@@ -179,6 +177,7 @@ var CustomElementDefinition;
       // 15:
       // @type {CustomElementDefinition}
       var definition = {
+        extends: options.extends,
```
treshugart: This is so we can check it later.

src/CustomElements/v1/CustomElements.js
```diff
@@ -440,6 +439,10 @@ var CustomElementDefinition;
         }
       }
     },
+      
+    _getDefinitionForNode: function(node) {
```
treshugart: Used in both `_addNodes()` and `_removeNodes()`.

tests/CustomElements/v1/js/upgrade.js
```diff
@@ -61,4 +61,32 @@ suite('Upgrades', function() {
     assert.instanceOf(e2, X2);
   });
 
+  suite('customised build-in elements', function() {
```
treshugart: It would be good to also test what happens if an invalid `{ is }` is passed.

Some other questions, do you think it would be prudent to add a test for every single possible built-in that can be extended here?

treshugart: I wasn't sure the best place to put these tests. Upgrade *seemed* relevant, but the tests inside it at first glance didn't correlate well with the tests I wanted to write. I just put them here for now. Can easily change later.

justinfagnani:
Thanks @treshugart! I'll take a close look at this asap

silentHoo:
How can we support here?

TimvdLippe:
@treshugart Very sorry, but I have to close this PR due to age. We will be working on this soon per https://github.com/webcomponents/custom-elements/issues/13 😢 Really sorry we left no response for such a long time.

treshugart:
No worries, @TimvdLippe, toted understand :)

PR #601, opened 2016-08-23
Update HTMLLabelElement.form test for new spec.

justinfagnani:
See: https://bugs.chromium.org/p/chromium/issues/detail?id=607868


azakus:
LGTM I guess

PR #609, opened 2016-09-01
Custom Style

azakus:


src/ShadyCSS/ShadyCSS.js
```diff
@@ -0,0 +1,290 @@
...
+        // update template
+        ApplyShim.transformRules(template._styleAst, is);
+        let target = this.nativeShadow ? template.content : null;
+        let placeholder = placeholderMap[is];
+        if (template._style) {
+          let s = template._style;
+          s.parentNode.removeChild(s);
+          template._style = null;
+        }
+        template._style = this._generateStaticStyle(host, template._styleAst, target, placeholder);
```
sorvell: We only need to update the textContent of the style in this case, isn't that so? Consider if it's worth the optimization to do that instead of replacing the style.

src/ShadyCSS/ShadyCSS.js
```diff
@@ -0,0 +1,290 @@
...
+        // update instance if native shadowdom
+        if (this.nativeShadow) {
+          let style = host.shadowRoot.querySelector('style');
+          style.parentNode.removeChild(style);
+          this._generateStaticStyle(host, template._styleAst, host.shadowRoot, placeholder);
```
sorvell: Same as comment above.

src/ShadyCSS/custom-style.js
```diff
@@ -0,0 +1,146 @@
...
+let customStyles = [];
+
+let hookFn = null;
+
+function enqueueDocumentValidation() {
+  if (enqueued) {
+    return;
+  }
+  enqueued = true;
+  if (window.HTMLImports) {
```
sorvell: Let's add a comment about the strategy here and acknowledge that there can be FOUC

src/ShadyCSS/custom-style.js
```diff
@@ -0,0 +1,146 @@
...
+
+function validateDocument() {
+  if (enqueued) {
+    ShadyCSS.updateStyles();
+    enqueued = false;
+  }
+}
+
+function CustomStyle() {
+  const self = HTMLElement.call(this);
```
sorvell: This doesn't work without the CE polyfill does it?

src/ShadyCSS/custom-style.js
azakus: Done above this

sorvell:
LGTM.

PR #618, opened 2016-09-14
Quick fix #617

### LINKED ISSUE #617: Throws on iOS 10
Including `webcomponents.js` v0.7.22 in Safari on iOS 10.0.1 throws

```
TypeError: undefined is not an object (evaluating 'descriptor.value')
installProperty — webcomponents.js:235
registerInstanceProperties — webcomponents.js:182
registerInternal — webcomponents.js:269
register — webcomponents.js:260
(anonymous function) — webcomponents.js:4365
global code — webcomponents.js:4367
```

The part in question is
```js
var descriptor = getDescriptor(source, name);
var getter, setter;
if (typeof descriptor.value === "function") {
```

and the error happens when `source` is the window object and `name` is `'showModalDialog'` (returns `undefined`).

It can be quick-fixed by changing `getDescriptor` in `ShadowDOM/wrappers.js` to:

```js
  function getDescriptor(source, name) {
    try {
      return Object.getOwnPropertyDescriptor(source, name) || dummyDescriptor;
    } catch (ex) {
      // JSC and V8 both use data properties instead of accessors which can
      // cause getting the property desciptor to throw an exception.
      // https://bugs.webkit.org/show_bug.cgi?id=49739
      return dummyDescriptor;
    }
  }
```

qwtel:


src/ShadowDOM/wrappers.js
```diff
@@ -182,7 +182,7 @@ window.ShadowDOMPolyfill = {};
 
   function getDescriptor(source, name) {
     try {
-      return Object.getOwnPropertyDescriptor(source, name);
+      return Object.getOwnPropertyDescriptor(source, name) || dummyDescriptor;
```
addyosmani: My expectation is this should throw an exception which we catch in the guard lower down. As the errors happens due to certain conditions with `window` and `showModalDialog`, is there a reason we wouldn't want to check for source === window && name === 'showModalDialog' and only early return the extra dummyDescriptor there?

tschwartz:
@qwtel Thoughts?

qwtel:
My assumption was that `showModalDialog` could be the first of many such errors. Turns out it's the only one. I've updated the PR.

src/ShadowDOM/wrappers.js
qwtel: My assumption was that showModalDialog could be the first of many such errors. Turns out it's the only one. I've updated the PR.

phpthug:
@addyosmani @ebidel Any additional feedback for @qwtel? As iOS10 becomes more widely used, this is going to start impacting more users.

mattppp:
Any update on this?

ArturKlajnerok:
Hey, @addyosmani @azakus @sorvell is there anything we could help / change to get this fix out ?
Since iOS 10 seems to have more than 40% market share this quite an issue  😸

jbwyme:
Is anyone actually actively maintaining this lib?

zanona:
Crazy isn't it. ¬¬

phpthug:
Any idea when this will make it into a tagged release?

samyue:
I have got this iphone ios10 issue as well. are there any ways for us to work around this issue before official release?

jannakha:
I had to use issue #617 fix and move ALL webcomponentsjs/* files into S3 to make it work, because webcomponents.js is not min'ed and not still pulls all refs dynamically 

@addyosmani please release ASAP, production is suffering

samyue:
@jannakha yeah, using the #617 fix to modify the getDescriptor function is working

PR #620, opened 2016-09-14
Rename parameters where parameter name is equal to function name

tuohis:
Fixes compatibility with older Webkits, especially Qt4.8's QtWebKit

A function with an equally named parameter crashes JavaScript execution in some versions of Webkit. A simple fix enables using Webcomponents.js in many embedded devices that use the older version of Qt and the flavor of Webkit that comes with it.

### DIFF
--- a/src/ShadowDOM/wrappers/Node.js
+++ b/src/ShadowDOM/wrappers/Node.js
@@ -632,19 +632,19 @@
       }
       return s;
     },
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Closing due to age and this has probably be fixed in the mean time. Especially as the polyfills are minified and this clash will likely never occur.

PR #624, opened 2016-09-19
Partial fix for #112

nazar-pc:
This is a partial fix for #112. It allows to select elements from element even when there is no `<content>` in its Shadow Root.
I know this a hack, not a complete fix, but:
- this is much better than nothing
- complete implementation would be quite complex (from what I understood)
- complete implementation is not likely to ever happen because of v1

Having that said I think it is better to have at least this partial fix.

### DIFF
--- a/src/ShadowDOM/querySelector.js
+++ b/src/ShadowDOM/querySelector.js
@@ -136,6 +136,7 @@
     var target = unsafeUnwrap(this);
     var list;
     var root = getTreeScope(this).root;
+    root = (!deep && root.shadowRoot) || root;
     if (root instanceof scope.wrappers.ShadowRoot) {
       // We are in the shadow tree and the logical tree is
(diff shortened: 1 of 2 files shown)


src/ShadowDOM/querySelector.js
```diff
@@ -136,6 +136,7 @@
     var target = unsafeUnwrap(this);
     var list;
     var root = getTreeScope(this).root;
+    root = (!deep && root.shadowRoot) || root;
```
gronke: I would suggest to merge this line into the line above. Would not be much harder to read, but saves calling getTreeScope in case the condition is true.

nazar-pc: I think it would be much harder to read, since `root` is used on both sides of `||`, also you'll need to call it anyway regardless of `deep` value.

PR #636, opened 2016-10-16
Event handler bug

jbwyme:
`target.addEventListener_(type, dispatchOriginalEvent, true)` is used to bind listeners to the target regardless if `useCapture` is true. That's ok until you go to remove:

```
if (found && count === 1) {
    var target = getTargetToListenAt(this);
    target.removeEventListener_(type, dispatchOriginalEvent, true);
}
```

This means if you bind an event listener using `useCapture: true`, bind another event listener using `useCapture: false`, and then remove the second listener it will stop listening to the target element even though we still expect to have the first event listener bound.

### DIFF
--- a/src/ShadowDOM/wrappers/events.js
+++ b/src/ShadowDOM/wrappers/events.js
@@ -777,7 +777,7 @@
       listeners.push(listener);
 
       var target = getTargetToListenAt(this);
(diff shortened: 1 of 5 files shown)


TimvdLippe:

PR #640, opened 2016-10-21
added support for WebKitMutationObserver

sfinktah:
Is there a reason you don't support WebKitMutationObserver?  It's the only MutationObserver available on my iPad which is only just short of current iOS (9.3.3 I believe), and appears to work fine.

re: guidelines, permission slips, etc, it's a 4 word change, if you count `global` as a changed word.  Seriously.

### DIFF
--- a/src/MutationObserver/MutationObserver.js
+++ b/src/MutationObserver/MutationObserver.js
@@ -567,9 +567,13 @@
   global.JsMutationObserver = JsMutationObserver;
 
   if (!global.MutationObserver) {
-    global.MutationObserver = JsMutationObserver;
-    // Explicltly mark MO as polyfilled for user reference.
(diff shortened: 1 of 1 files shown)


sfinktah:
You, your CI is broken.

ebidel:
The unprefixed `MutationObserver` has been around since Safari 6.1 and iOS Safari 7.1. Those are over two years old. http://caniuse.com/#feat=mutationobserver

iOS 9 definitely has the unprefixed version.

sfinktah:
I totally agree that **apparently** iOS 9 has MutationObserver.  But my iPad running iOS 9.3.3 disagrees.  iOS 10 is only a few months old really.

TimvdLippe:
We no longer polyfill this feature, so this PR has sadly aged.

PR #644, opened 2016-10-28
fixed memory leak while importing html elements in IE

bbsteventill:
541 - fixed ie memory leak when importing multiple imports

```
<!-- import.html -->

<link rel="import" href="/assets/paper-something/paper-something.html" />

<!-- index.html -->

<html><head></head><body>
<link rel="import" href="/import.html" />
</body></html>
```

this wil no longer leak in IE 11

### DIFF
--- a/src/CustomElements/observe.js
+++ b/src/CustomElements/observe.js
@@ -252,9 +252,8 @@ function takeRecords(node) {
     node = node.parentNode;
   }
 
-  // this needs to be on head or it will leak in IE
-  // IE does not like it when you have non-standard attributes on root dom's, so put
-  // the observer on the head element
+  // The node is a ShadowRoot, an IE will have a memory leak if you put the observer
+  // directly on the ShadowRoot, so put it on the head so it does not leak
   var observer = node.head.__observer;
   if (observer) {
     handler(node, observer.takeRecords());
@@ -263,7 +262,6 @@ function takeRecords(node) {
 }
 
 var forEach = Array.prototype.forEach.call.bind(Array.prototype.forEach);
-
 
 // observe a node tree; bail if it's already being observed.
 function observe(inRoot) {
@@ -276,8 +274,9 @@ function observe(inRoot) {
   // Give the handler access to the root so that an 'in document' check can
   // be done.
 
-  // IE requires that you put an observer on child elements of the DOM or it will leak
-  // at this point inRoot == #document
+  // originally the observer was on the ShadowRoot (inRoot) (single observer); 
+  // this causes a memory leak within IE.  To fix this, we must put a an observer
+  // on both the head and body nodes on the ShadowRoot
   var observer = new MutationObserver(handler.bind(this, inRoot));
   observer.observe(inRoot.head, {childList: true, subtree: true});
   observer.observe(inRoot.body, {childList: true, subtree: true});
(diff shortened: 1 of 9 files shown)


justinfagnani:
@bbsteventill can you give a brief description of the fix and why it prevents the memory leak?

src/CustomElements/observe.js
```diff
@@ -263,16 +267,26 @@ var forEach = Array.prototype.forEach.call.bind(Array.prototype.forEach);
...
   // garbage collected once all references to the `inRoot` node are gone.
   // Give the handler access to the root so that an 'in document' check can
   // be done.
+
+  // IE requires that you put an observer on child elements of the DOM or it will leak
+  // at this point inRoot == #document
   var observer = new MutationObserver(handler.bind(this, inRoot));
-  observer.observe(inRoot, {childList: true, subtree: true});
-  inRoot.__observer = observer;
+  observer.observe(inRoot.head, {childList: true, subtree: true});
```
bbsteventill: that my friend is what will causes a memory leak in IE; which is what it was before

src/CustomElements/observe.js
```diff
@@ -251,7 +251,11 @@ function takeRecords(node) {
   while (node.parentNode) {
     node = node.parentNode;
   }
-  var observer = node.__observer;
+
+  // this needs to be on head or it will leak in IE
+  // IE does not like it when you have non-standard attributes on root dom's, so put
```
bbsteventill: My findings are attributes that are not defined in the w3 spec.  I found an example that will cause a leak  as part of the UrlResolver in polymer where a 'a' element is being appended to the root document, this breaks, but if you set the baseURI that does not cause a memory leak.  I think it is because baseURI is defined in the w3 spec while __url is not (https://github.com/Polymer/polymer/pull/3964/files)

a root dom, is any root document that has been created with document.implementation.createHTMLDocument

this is why we need the node.head && node.head.__observer, because placing the observer directly on the node (this is where the code is poorly named because node is always a root shadow document at this point in the code)

we do have test coverage of this, the tests will break if you remove the createElement('head') && createElement('body) that was added Element.js

jdmedina:
This merge causes an error in IE11 for my components:

SCRIPT5007: Unable to get property 'getAttribute' of undefined or null reference
webcomponents-lite.min.js, Line: 64, Column: 3

`  var src = script.getAttribute('src');`

bbsteventill:
@jdmedina More information is required; first do not use your own custom minified version of the script.  Second I know that the debugger in IE does different things then what happens in chrome; for example when you try to view properties on a prototype I think the IE debugger evaluates the properties for you; this may be bad because things like script.getAttribute('src'), your script object may be null

jdmedina:
@bbsteventill By custom you mean the lite version (sans shady dom)? Even though IE was saying the error is on that minified file, it was not. I am using Fiddler to hot-swap the file we're using in our test server with the web components-lite.js in the master. 

Even with the debugger off, the components break on my page and don't work.

Let me know what other type of extra info you need.

jay8t6:
@justinfagnani @bbsteventill when is this gonna be tagged?

omsmith:
Seeing very big improvements in IE11 with this PR. Thanks very much @bbsteventill :)

PR #655, opened 2016-12-09
Only dispatch load event when import was successful

gronke:
When an import was not successfully loaded in first place, the load event should not be dispatched on the element if trying to import a second time. This change checks there is actually content loaded before invoking the load event.

Tests were added according to the [notes in the webcomponents import spec](https://w3c.github.io/webcomponents/spec/imports/#fetching-import)

### DIFF
--- a/src/HTMLImports/parser.js
+++ b/src/HTMLImports/parser.js
@@ -116,7 +116,7 @@ var importParser = {
     }
     this.markParsingComplete(elt);
     // fire load event
-    if (elt.__resource && !elt.__error) {
+    if (elt.__resource && !elt.__error && elt.import) {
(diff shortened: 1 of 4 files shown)


gronke:
@cdata this fix is required for the referenced refactoring branch that adds similar tests there as well. One of the tests added here fails on master and shows why load should not be fired when link.import is undefined/null.

tomalec:
I have noticed there is a new repo with HTML Imports polyfill https://github.com/webcomponents/html-imports.
Maybe we should start PRing there?

TimvdLippe:
Closing this, as v2 no longer contains HTML imports. Sorry for the long wait 😢

gronke:
No worries @TimvdLippe. Thanks for the update!

PR #661, opened 2017-01-12
Gulp all the things

notwaldorf:
🍍 `Polymer` tests with this branch: https://github.com/Polymer/polymer/pull/4220
🍍 `Paper-dialog` tests with this branch: https://github.com/PolymerElements/paper-dialog/pull/139

### 👀  Changes:
- there are 4 bundles: `hi`, `hi-ce`, `hi-ce-sd`, and `hi-ce-sd-pf` also know as "everything and the kitchen sync", which is `webcomponents-lite.js`
- `webcomponents-lite.min.js#v0` was ` 118 KB`; this `v1` seems to be `106 KB`.  
- there is a `webcomponents-loader.js` that tries to do the smart thing and async load you the right polyfill
- the tests use this loader, unless they're tests that check that flags load, in which case they use `webcomponents-lite.js`
- _only_ the`webcomponents-lite.js` bundle handles flags
- copied the old `document.write` `webcomponents-lite.js` to `webcomponents-debug.js`
- ~~started using the `v1` branch of `HTMLImports` :tada:~~ (too many problems with that branch rn)
- some tests were dumb and needed fixes. more tests are probably still dumb.

### 😭  Hacks that are needed because of bugs in dependences:
- there's a `webcomponents-none.js` that Chrome needs to load, because we still need to load `shadyCSS.js`. This will be removed once the Shady CSS bits of code are moved into Polymer

/cc @azakus @sorvell @justinfagnani


gulpfile.js
```diff
@@ -1,137 +1,72 @@
 /**
  * @license
- * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
cdata: :two: :zero: :one: :seven:

notwaldorf: 😱

entrypoints/webcomponents-hi-ce-sd-index.js
```diff
@@ -0,0 +1,23 @@
...
+/*
+ * Polyfills loaded: HTML Imports, Custom Elements, Shady DOM/Shady CSS
+ * Used in: Safari 9, Firefox, Edge
+ */
+
+import '../bower_components/html-imports/src/html-imports.js'
+import '../src/pre-polyfill.js'
+import '../bower_components/custom-elements/custom-elements.min.js'
+import '../bower_components/shadydom/shadydom.min.js'
+import '../bower_components/shadycss/src/ShadyCSS.js'
```
justinfagnani: temporary here too

notwaldorf: Is it? this is `hi-ce-sd` which needs shady, no?

notwaldorf:
IE11 is broken: nobody is surprised.

webcomponents-lite-debug.js
```diff
@@ -0,0 +1,76 @@
...
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+(function() {
```
cdata: I'm afraid of calling this the "debug" version of "lite", since it uses a completely different strategy to load in the scripts. `document.write` will cause all of the loaded scripts to be evaluated in the order that they are written, while the new "lite" strategy causes scripts to be evaluated in the order they load. An error caused by unexpected loading order in "lite" will actually be obscured by this "debug" version.

webcomponents-lite.js
```diff
@@ -48,30 +47,36 @@
...
+  }
+  if (!('attachShadow' in Element.prototype)) {
+    polyfills.push('sd');
+  }
+  if (!('content' in document.createElement('template')) || !window.Promise || !window.URL) {
+    polyfills.push('pf');
+  }
 
+  // TODO(notwaldorf): This is a temporary hack because of pre/post-polyfill.js,
+  // and needs to not exist at all.
```
cdata: 👍 Don't forget to remove this :)

webcomponents-lite-debug.js
cdata: Oh, nevermind. I keep forgetting that the new strategy still bundles scripts. Please disregard!

webcomponents-lite.js
notwaldorf: I can't remove it until there's changes in polymer (it also needs to load ShadyCSS right now)

package.json
```diff
@@ -21,13 +21,13 @@
   },
   "homepage": "http://webcomponents.org",
   "devDependencies": {
+    "babel-preset-babili": "0.0.9",
     "gulp": "^3.8.8",
-    "gulp-audit": "^1.0.0",
-    "gulp-concat": "^2.4.1",
-    "gulp-header": "^1.1.1",
-    "gulp-rename": "^1.2.2",
```
bicknellr: btw, `gulp-rename` is still used but isn't in the new set of deps.

notwaldorf: Fixed! Thanks!

webcomponents-loader.js
```diff
@@ -0,0 +1,76 @@
...
+        // the document becomes interactive, so wait for the correct state change.
+        if (document.readyState !== 'interactive') {
+          function once() {
+            document.removeEventListener('readystatechange', once);
+            window.HTMLImports.whenReady(callback);
+          }
+          document.addEventListener('readystatechange', once);
+        } else {
+          requestAnimationFrame(function() {
+            callback();
```
valdrinkoshi: This should return the `importInfo` with loaded/errored imports https://github.com/webcomponents/html-imports/blob/cleanup/src/html-imports.js#L735-L741

notwaldorf: Ugh, fake-polyfilling an API that comes from a polyfill is a path down a dark alley, in my opinion. This is to fix the problem that a lot of code (including our friend `browser.js` relies on this callback being called), but if your use case is more advanced than that (and I don't see anyone using `importInfo`), then not sure we should help in that case.

notwaldorf:
@cdata @dfreedm I think this is ready for another look!

webcomponents-loader.js
valdrinkoshi: Updated `html-imports#cleanup` to invoke the callback w/o any arguments 👌

entrypoints/webcomponents-hi-ce-sd-index.js
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+ * Polyfills loaded: HTML Imports, Custom Elements, Shady DOM/Shady CSS
+ * Used in: Safari 9, Firefox, Edge
+ */
+
+import '../bower_components/html-imports/src/html-imports.js'
+import '../bower_components/custom-elements/custom-elements.min.js'
+import '../bower_components/shadydom/shadydom.min.js'
+import '../bower_components/shadycss/src/ShadyCSS.js'
```
bicknellr: BTW, I just heard from @azakus that ShadyCSS comes in two parts, the file included here and `custom-style-element.js`.

notwaldorf:
OMG!!! With both custom-elements and html-imports experimental branches, `Polymer` tests are ✅ and `paper-dialog` tests are as ✅ as the base branch!! 🎉

robdodson:
can I feature request a `ce-sd` bundle here?

notwaldorf:
@rob if you're using this for just vanilla custom elements, is there a reason why you wouldn't just do
```
import '../bower_components/html-imports/html-imports.min.js'
import '../bower_components/custom-elements/custom-elements.min.js'
import '../bower_components/shadydom/shadydom.min.js'
import '../bower_components/shadycss/shadycss.min.js'
```

We're not doing anything else on top of it that you would need without polymer

valdrinkoshi:
@notwaldorf looks like firefox doesn't like when you steal focus from it when tests are running (at least that's how i can repro it locally)... 😭

webcomponents-loader.js
```diff
@@ -0,0 +1,76 @@
...
+    polyfills.push('hi');
+  }
+
+  // Stub out HTMLImports if we're using native imports
+  window.HTMLImports = {
+    whenReady: function(callback) {
+      if (useNativeImports) {
+        // When native imports boot, the are "ready" the first rAF after
+        // the document becomes interactive, so wait for the correct state change.
+        if (document.readyState !== 'interactive') {
```
valdrinkoshi: Will this invoke the callback if `readyState` is already `complete`?

README.md
```diff
@@ -3,48 +3,44 @@ webcomponents.js
...
+- **Shadow DOM**: provides encapsulation by hiding DOM subtrees under shadow roots ([spec](https://w3c.github.io/webcomponents/spec/shadow/)).
 
 ## Releases
 
-Pre-built (concatenated & minified) versions of the polyfills are maintained in the [tagged versions](https://github.com/webcomponents/webcomponentsjs/releases) of this repo. There are two variants:
-
-`webcomponents.js` includes all of the polyfills.
-
-`webcomponents-lite.js` includes all polyfills except for shadow DOM.
+Pre-built (concatenated & minified) versions of the polyfills are maintained in the [tagged versions](https://github.com/webcomponents/webcomponentsjs/releases) of this repo. There are several variants:
```
sorvell: Please add a few notes about manually loading the polyfills. Perhaps we can say that (1) load order matters (shadydom before custom-elements before shadycss) and (2) that users can use the files in `entrypoints` and the code in `webcomponents-loader.js` as a guide?

README.md
```diff
@@ -74,7 +71,7 @@ Copyright (c) 2015 The Polymer Authors. All rights reserved.
 
 Under native HTML Imports, `<script>` tags in the main document block the loading of such imports. This is to ensure the imports have loaded and any registered elements in them have been upgraded.
 
-The webcomponents.js and webcomponents-lite.js polyfills parse element definitions and handle their upgrade asynchronously. If prematurely fetching the element from the DOM before it has an opportunity to upgrade, you'll be working with an `HTMLUnknownElement`.
+The `webcomponents-lite.js` polyfill parse element definitions and handle their upgrade asynchronously. If prematurely fetching the element from the DOM before it has an opportunity to upgrade, you'll be working with an `HTMLUnknownElement`.
 
 For these situations (or when you need an approximate replacement for the Polymer 0.5 `polymer-ready` behavior), you can use the `WebComponentsReady` event as a signal before interacting with the element. The criteria for this event to fire is all Custom Elements with definitions registered by the time HTML Imports available at load time have loaded have upgraded.
```
sorvell: Let's remove the Polymer 0.5 reference from here.

entrypoints/webcomponents-hi-ce-index.js
```diff
@@ -0,0 +1,23 @@
...
+'use strict';
+
+/*
+ * Polyfills loaded: HTML Imports, Custom Elements
+ * Used in: Safari 10, Firefox once SD is shipped
+ */
+
+import '../bower_components/html-imports/src/html-imports.js'
+import '../bower_components/custom-elements/custom-elements.min.js'
+//TODO(notwaldorf): this is temporary and should be removed.
```
sorvell: Let's make it more clear that the next 2 lines should be removed (via spacing?) and say it's blocked on https://github.com/webcomponents/shadycss/issues/46.

entrypoints/webcomponents-hi-index.js
```diff
@@ -0,0 +1,22 @@
...
+*/
+'use strict';
+
+/*
+ * Polyfills loaded: HTML Imports
+ * Used in: Safari Tech Preview
+ */
+
+import '../bower_components/html-imports/src/html-imports.js'
+//TODO(notwaldorf): this is temporary and should be removed.
```
sorvell: same as previous comment

entrypoints/webcomponents-none-index.js
```diff
@@ -0,0 +1,22 @@
...
+'use strict';
+
+/*
+ * Polyfills loaded: None
+ * Used in: Chrome
+ */
+
+// TODO: This needs to not exist at all.
+
+//TODO(notwaldorf): this is temporary and should be removed.
```
sorvell: same as previous comment

webcomponents-loader.js
```diff
@@ -0,0 +1,77 @@
...
+          function once() {
+            document.removeEventListener('readystatechange', once);
+            window.HTMLImports.whenReady(callback);
+          }
+          document.addEventListener('readystatechange', once);
+        } else {
+          requestAnimationFrame(function() {
```
sorvell: Let's add a TODO here that ideally this would return synchronously if imports were not loading. Let's make an issue for this as well.

webcomponents-loader.js
```diff
@@ -0,0 +1,77 @@
...
+    polyfills.push('sd');
+  }
+  if (!window.customElements || window.customElements.forcePolyfill) {
+    polyfills.push('ce');
+  }
+  if (!('content' in document.createElement('template')) || !window.Promise || !window.URL) {
+    polyfills.push('pf');
+  }
+
+  // TODO(notwaldorf): This is a temporary hack because Chrome still needs to
```
sorvell: Let's make an issue for this for tracking.

webcomponents-loader.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  if (polyfills.length) {
+    var script = document.querySelector('script[src*="webcomponents-loader.js"]');
+    let newScript = document.createElement('script');
+    // Load it from the right place.
+    var url = script.src.replace(
+        'webcomponents-loader.js', `webcomponents-${polyfills.join('-')}.js`);
+    newScript.src = url;
+    console.log('Loaded: ', newScript.src);
```
sorvell: Remove?

src/post-polyfill.js
```diff
@@ -18,4 +18,8 @@
     });
   });
 
+  if (customElements && customElements.polyfillWrapFlushCallback) {
+    customElements.polyfillWrapFlushCallback(flush => HTMLImports.whenReady(flush));
```
gorshkov-leonid: value: function polyfillWrapFlushCallback(outer) {
          this._documentConstructionObserver.disconnect();
          ....
 }

 _createClass(DocumentConstructionObserver, [{
        key: 'disconnect',
        value: function disconnect() {
          this._observer.disconnect();  //this._observer == undefined
        }
      }
sorry for the span, but I do not understand polyfillWrapFlushCallback can be called before documentReady?

webcomponents-loader.js
```diff
@@ -0,0 +1,81 @@
...
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+(function() {
+  // Feature detect which polyfill needs to be imported.
+  let polyfills = [];
```
gorshkov-leonid: may be using var for better compatibility

webcomponents-loader.js
```diff
@@ -0,0 +1,81 @@
...
+  let useNativeImports = ('import' in document.createElement('link'));
```
gorshkov-leonid: may be using var for better compatibility

webcomponents-loader.js
```diff
@@ -0,0 +1,81 @@
...
+  } else if (polyfills.length === 4) {  // hi-ce-sd-pf is actually called lite.
+    polyfills = ['lite'];
+  }
+
+  if (polyfills.length) {
+    var script = document.querySelector('script[src*="webcomponents-loader.js"]');
+    let newScript = document.createElement('script');
+    // Load it from the right place.
+    var url = script.src.replace(
+        'webcomponents-loader.js', `webcomponents-${polyfills.join('-')}.js`);
```
gorshkov-leonid: may be using simple concatination instead string interpolation for better compatibility?

webcomponents-hi-sd-ce.js
```diff
@@ -1,6 +1,6 @@
...
...rkingURL)return new URL(ro,no).href;let so=Ft.__tempDoc;return so||(so=document.implementation.createHTMLDocument('temp'),Ft.__tempDoc=so,so.__base=so.createElement('base'),so.head.appendChild(so.__base),so.__anchor=so.createElement('a')),so.__base.href=no,so.__anchor.href=ro,so.__anchor.href||ro}},Ut={async:!0,load(ro){return new Promise((no,so)=>{if(!ro)so({resource:'error: href must be specified'});else if(ro.match(/^data:/)){const ao=ro.split(','),lo=ao[0];let io=ao[1];io=-1<lo.indexOf(';base64')?atob(io):decodeURIComponent(io),no({resource:io})}else{const ao=new XMLHttpRequest;ao.open('GET',ro,Ut.async),ao.addEventListener('readystatechange',()=>{if(4===ao.readyState){let lo;try{const po=ao.getResponseHeader('Location');po&&(lo='/'===po.substr(0,1)?location.origin+po:po)}catch(po){console.error(po.message)}const io={resource:ao.response||ao.responseText,redirectedUrl:lo};304===ao.status||0===ao.status||200<=ao.status&&300>ao.status?no(io):so(io)}}),ao.send()}})}},Wt=/Trident/.test(navigator.userAgent)||/Edge\/\d./i.test(navigator.userAgent),qt='onunhandledrejection'in window,Bt='link[rel=import]',jt='import-disable',Gt=`link[rel=stylesheet][href][type=${jt}]`,Xt=`${Bt}, ${Gt},
```
gorshkov-leonid: =>, let, const... - not compatible with Firefox, IE 11 (SyntaxError: let is a reserved identifier / SCRIPT1002: Syntax error ): 
may be compile all to honest es5?

notwaldorf: @azakus has a branch that does exactly that, using closure. Please have patience; this PR is still in development, isn't the final PR for the `v1` brach, and depends on a lot of other PRs, so reviewing it right now isn't entire useful :)

PR #663, opened 2017-01-19
Remove URL polyfill

notwaldorf:
Depends on https://github.com/webcomponents/webcomponentsjs/pull/661

Once `webcomponents/html-imports#cleanup` is merged, `HTMLImports` won't use a  constructable URL, so we no longer need a polyfill for it.


notwaldorf:
Going to close this and open the new one, because changing the base branch sounds daunting

PR #665, opened 2017-01-26
Revert "fixed memory leak while importing html elements in IE"

sorvell:
Reverts webcomponents/webcomponentsjs#644.

Unfortunately, this PR is incorrect and must be reverted. ShadowRoot elements do not contain `<head>` or `<body>` elements as was [added here](https://github.com/webcomponents/webcomponentsjs/commit/d29f3aca32a65d94ad98961b8ed21ba73ee6935a#diff-cbc0ec9185c39ec81f72076daa06b4d3R79). The custom elements polyfill was [changed](https://github.com/webcomponents/webcomponentsjs/commit/a2382fd6e095ee3ac1479786d8acf0f29a9c57f7#diff-1883af06f66c8c928d72352cd1c8e0b4R274) to always observe `<head>` and `<body>` on every element it observes, including documents and shadowRoots. This caused any custom elements added to shadowRoots to not be seen by the mutation observer since the observer is observing disconnected, unused `shadowRoot.head` and `shadowRoot.body` elements rather than the `shadowRoot` itself.

We'll make sure the issue here gets re-opened and will definitely consider other fixes for the core issue.

### DIFF
--- a/src/CustomElements/observe.js
+++ b/src/CustomElements/observe.js
@@ -251,10 +251,7 @@ function takeRecords(node) {
   while (node.parentNode) {
     node = node.parentNode;
   }
-
-  // The node is a ShadowRoot, an IE will have a memory leak if you put the observer
-  // directly on the ShadowRoot, so put it on the head so it does not leak
-  var observer = node.head.__observer;
+  var observer = node.__observer;
   if (observer) {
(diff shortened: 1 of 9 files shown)


kevinpschaaf:
LGTM, failing tests seem to be status quo.

PR #669, opened 2017-02-02
Copy custom event properties to IE Event shim

ostap0207:
WebComponents Event shim has a conflict with [PrototypeJs Event
object][1]. WebComponents Event shim create a new Event, which clears own properties of PrototypeJS
Event object.

This can be fixed by copying own properties of original Event to the shim Event.

[1]: http://api.prototypejs.org/dom/Event/


ostap0207:
Already fixed

PR #671, opened 2017-02-04
Refresh bower packages on gulping

azakus:
Add `debug` gulp task to not refresh bower


notwaldorf:
LGTM. Tests be red bc base tests be red

PR #677, opened 2017-02-10
Use html-imports-native. Fire WebComponentsReady once.

valdrinkoshi:
Instead of stubbing `HTMLImports` in here, depend on the native/lighter implementation available on https://github.com/webcomponents/html-imports/pull/29.

`WebComponentsReady` would fire twice when using `webcomponents-loader.js`:
- one is fired by [`webcomponents-loader.js` itself](https://github.com/webcomponents/webcomponentsjs/blob/v1/webcomponents-loader.js#L76)
- another is fired by any bundle that loads [`post-polyfill.js`](https://github.com/webcomponents/webcomponentsjs/blob/v1/src/post-polyfill.js#L48), e.g. [hi-ce bundle](https://github.com/webcomponents/webcomponentsjs/blob/v1/entrypoints/webcomponents-hi-ce-index.js#L25)

My assumption is that `WebComponents` is just an assembler of polyfills, and as a user I want to rely on having the object `WebComponents` defined, and `WebComponentsReady` event to be fired. Whatever is done by `webcomponents-none.js` should eventually be handled by `webcomponents-loader.js`.


valdrinkoshi:
@kevinpschaaf updated the description 👌

notwaldorf:
So in the likely case that `webcomponents-none` can never go away, we are definitely firing this event twice (just noticed it in a demo) :(

valdrinkoshi:
Moved the fix for double firing of `WebComponentsReady` in this PR https://github.com/webcomponents/webcomponentsjs/pull/688.

PR #680, opened 2017-02-10
Cleanup readme and add sd-ce bundle

### LINKED ISSUE #678: v1 readme needs cleanup
In particular, there's references about `currentDocument` not working, which isn't true.

notwaldorf:
Fixes #678. 
@robdodson for 🎁 
@azakus for 👀 

Dan: I don't know enough about the "other limitation" sections -- there's 2 about shady CSS that look like they might still be a problem in v1, but wasn't 100% sure so I left them.


robdodson:
yay! thank you :)

PR #690, opened 2017-02-16
fix ShadyCSS entrypoint, load ScopingShim

azakus:


webcomponents-loader.js
```diff
@@ -43,12 +38,9 @@
     document.head.appendChild(newScript);
   }
   // Ensure `WebComponentsReady` is fired also when there are no polyfills loaded.
-  // TODO(valdrin): only check for `!polyfills.length` once 'none' bundle
-  // is removed. Addressing this is blocked on
-  // https://github.com/webcomponents/shadycss/issues/46.
-  if (polyfills[0] === 'none') {
+  if (!polyfills.length) {
```
valdrinkoshi: nit: this could just be: 
```js
  document.head.appendChild(newScript);
} else {
  // Ensure `WebComponentsReady` is fired also when there are no polyfills loaded.
  requestAnimationFrame(function () {

PR #693, opened 2017-02-23
 Use better words about the bundles and the loader

notwaldorf:
I cleaned up the "Releases" section into a (maybe clearer) "How to use section", since we have a lot a build artifacts now and they're a bit intimidating.

👀  @kevinpschaaf

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,27 +3,54 @@ webcomponents.js
 
 [![Build Status](https://travis-ci.org/webcomponents/webcomponentsjs.svg?branch=master)](https://travis-ci.org/webcomponents/webcomponentsjs)
 
-A suite of polyfills supporting the [Web Components](http://webcomponents.org) specs.
+A suite of polyfills supporting the [Web Components](http://webcomponents.org) specs:
 
 - **Custom Elements**: allows authors to define their own custom tags ([spec](https://w3c.github.io/webcomponents/spec/custom/)).
 - **HTML Imports**: a way to include and reuse HTML documents via other HTML documents ([spec](https://w3c.github.io/webcomponents/spec/imports/)).
 - **Shadow DOM**: provides encapsulation by hiding DOM subtrees under shadow roots ([spec](https://w3c.github.io/webcomponents/spec/shadow/)).
 
-## Releases
+## How to use
 
-Pre-built (concatenated & minified) versions of the polyfills are maintained in the [tagged versions](https://github.com/webcomponents/webcomponentsjs/releases) of this repo. There are several variants:
+The polyfills are built (concatenated & minified) into several bundles that target
+different browsers and spec readiness:
 
-- `webcomponents-lite.js` includes all of the polyfills.
-- `webcomponents-loader.js` is a custom loader that dynamically load a minified polyfill
-bundle, using feature detection. The bundles that can be loaded are:
-  - `webcomponents-hi` -- HTML Imports (needed by Safari Tech Preview)
-  - `webcomponents-hi-ce` -- HTML Imports and Custom Elements (needed by Safari 10)
-  - `webcomponents-hi-ce-sd` -- HTML Imports, Custom Elements and Shady DOM/CSS (needed by Safari 9, Firefox, Edge)
-  - `webcomponents-lite` -- HTML Imports, Custom Elements, Shady DOM/CSS and generic platform polyfills (such as Template, ES6 Promise, Constructable events, etc.) (needed by Internet Explorer 11)
+- `webcomponents-hi` -- HTML Imports (needed by Safari Tech Preview)
+- `webcomponents-hi-ce` -- HTML Imports and Custom Elements (needed by Safari 10)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,27 +3,54 @@ webcomponents.js
...
+<head>
+  <script src="bower_components/webcomponentsjs/webcomponents-loader.js"></script>
+  <script>
+    // When the polyfill has loaded, lazy-load another custom element (otherwise,
+    // if we do this before CustomElements has loaded, the element registration
+    // will fail.)
+    window.addEventListener('WebComponentsReady', function() {
+      var s = document.createElement('script');
+      s.src = 'lazy-element.js';
+      document.head.appendChild(s);
```
kevinpschaaf: Hmmm... not sure the example is showing a particularly idiomatic use case (dynamically appending script after WCR?), thus might be hard to understand why WCR is important.  How about:

```html
<!-- Load polyfills; note that "loader" will these will load async -->
<script src="bower_components/webcomponentsjs/webcomponents-loader.js"></script>

<!-- Load a custom element definition via HTMLImports -->
<link rel="import" href="my-element.html">

<!-- Use the custom element -->
<my-element></my-element>

<!-- Interact with the upgraded element -->
<script>
  // Wait until polyfills (if any) have loaded before interacting with elements
  // Note this is only required for main-document or async scripts (not scripts in HTMLImports)
  window.addEventListener('WebComponentsReady', function() {
    let element = document.querySelector('my-element');
    let MyElement = customElements.get('my-element');
    console.assert(element instanceof MyElement);  // 👍
    element.someAPI(); // 👍
  });
</script>
```

README.md
```diff
@@ -3,27 +3,54 @@ webcomponents.js
...
-bundle, using feature detection. The bundles that can be loaded are:
-  - `webcomponents-hi` -- HTML Imports (needed by Safari Tech Preview)
-  - `webcomponents-hi-ce` -- HTML Imports and Custom Elements (needed by Safari 10)
-  - `webcomponents-hi-ce-sd` -- HTML Imports, Custom Elements and Shady DOM/CSS (needed by Safari 9, Firefox, Edge)
-  - `webcomponents-lite` -- HTML Imports, Custom Elements, Shady DOM/CSS and generic platform polyfills (such as Template, ES6 Promise, Constructable events, etc.) (needed by Internet Explorer 11)
+- `webcomponents-hi` -- HTML Imports (needed by Safari Tech Preview)
+- `webcomponents-hi-ce` -- HTML Imports and Custom Elements (needed by Safari 10)
+- `webcomponents-hi-sd-ce` -- HTML Imports, Custom Elements and Shady DOM/CSS (needed by Safari 9, Firefox, Edge)
+- `webcomponents-sd-ce` -- Custom Elements and Shady DOM/CSS (no HTML Imports)
+- `webcomponents-lite` -- all of the polyfills: HTML Imports, Custom Elements, Shady DOM/CSS and generic platform polyfills (such as Template, ES6 Promise, Constructable events, etc.) (needed by Internet Explorer 11)
```
kevinpschaaf: Add `.js` extensions so it's clear these are filenames?

README.md
```diff
@@ -3,27 +3,54 @@ webcomponents.js
...
+
+If you are only targetting a specific browser, you can just use the bundle that's
+needed by it; alternatively, if you're using server-side rendering, you can
+send the polyfill bundle that's necessary for the browser making that request.
+
+## `webcomponents-loader.js`
+
+Alternatively, this repo also comes with `webcomponents-loader.js`, a client-side
+loader that dynamically loads the correct polyfill bundle, using feature detection.
```
kevinpschaaf: s/correct polyfill bundle/minimum polyfill bundle

README.md
kevinpschaaf: Yikes, just realized that we might need to tweak loader so that `webcomponents-lite` is loaded for Edge too -- let's double-check tomorrow.  Technically Edge has Template but their [impl is pretty busted](https://github.com/webcomponents/template/commit/b6137231f28449bb66837beac82145b0ac9c069e) (we fix it up in the Template polyfill).  Since most testing is using `webcomponents-lite` this hasn't come up in tests.  😩

kevinpschaaf: Yeah... I think we need to add this check, and if it returns false need to load template polyfill (`pf` bundle):

```
(document.createDocumentFragment().cloneNode() instanceof DocumentFragment)
```

README.md
notwaldorf: Oh yeah that's waaaay better. Thanks!

notwaldorf:
@kevinpschaaf copy pasted all your changes, basically :)

README.md
kevinpschaaf: We merged this in https://github.com/webcomponents/webcomponentsjs/pull/694/files#diff-dba880ae0f7227822f1bb5a55896b463R24

notwaldorf:
Thanks!

PR #698, opened 2017-02-28
Use closure for bundles

azakus:
Simple builds for now, will advanced compile later

Mark built files as binary to hide generated changes


.gitattributes
```diff
@@ -0,0 +1,3 @@
+/webcomponents*.js binary
+/webcomponents*.js.map binary
```
notwaldorf: 🙌

package.json
```diff
@@ -21,17 +21,16 @@
...
-    "gulp-babel": "^6.1.2",
     "gulp-rename": "^1.2.2",
-    "gulp-sourcemaps": "1.9.0",
+    "gulp-sourcemaps": "^2.4.1",
     "rollup-stream": "^1.14.0",
     "run-sequence": "^1.2.2",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0",
-    "web-component-tester": "^6.0.0-prerelease.4"
+    "web-component-tester": "^6.0.0-prerelease.5"
```
notwaldorf: If you do `"web-component-tester": "Polymer/web-component-tester#^6.0.0-prerelease.6",` this should autowork maybe? Also we're on `.6` now, surprise.

azakus: looks like `.6` doesn't have an npm release yet

PR #704, opened 2017-03-03
Make everything advance compile

azakus:


src/post-polyfill.js
```diff
@@ -8,35 +8,38 @@
...
       if (flushCallback) {
         let cb = flushCallback;
         flushCallback = null;
         cb();
         return true;
       }
     }
-    let origWhenReady = HTMLImports.whenReady;
-    customElements.polyfillWrapFlushCallback(function(cb) {
+    let origWhenReady = HTMLImports['whenReady'];
```
notwaldorf: s/let/var/

src/pre-polyfill.js
```diff
@@ -13,20 +13,21 @@
   'use strict';
 
   // Establish scope.
-  window.WebComponents = window.WebComponents || {flags:{}};
+  window['WebComponents'] = window['WebComponents'] || {'flags':{}};
```
notwaldorf: is this because of strict?

azakus: No, this is because closure will rename properties that aren't quoted.

PR #705, opened 2017-03-03
Remove webcomponents-debug.js, make `gulp debug` be good enough

azakus:
webcomponents-debug required all polyfills to be installed, which is not
what we require for bower installations.

Instead, webcomponents-dev is all polyfills rolled into one, ES6,
without sourcemaps, for the easiest debug experience.


notwaldorf:
LGTM! 🚢

valdrinkoshi:
no `webcomponents-dev` bundle added?

PR #711, opened 2017-03-06
Make bundles with the Custom Elements es5 compatibility shim

### LINKED ISSUE #706: v1 - webcomponentsloader.js gives Invalid Character error in IE11
var url = script.src.replace(
      'webcomponents-loader.js', `webcomponents-${polyfills.join('-')}.js`);
    newScript.src = url;

Template literal string not supported in IE11.

### LINKED ISSUE #709: Add native-shim documentation to the README

### LINKED ISSUE #710: Add webcomponents-loader-es5 to load the native-shim on Browsers with native Custom Elements

azakus:
Fixes #710
Fixes #709

Also, fix loaders to be compatible with ES5-only browsers

Fixes #706


webcomponents-ce-es5.js
```diff
@@ -0,0 +1,267 @@
+(function () {
+'use strict';
+
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
```
notwaldorf: 2017 (sorry)

webcomponents-ce-es5.js
```diff
@@ -0,0 +1,267 @@
...
+ *  Compiling valid class-based custom elements to ES5 will satisfy these
+ *  requirements with the latest version of popular transpilers.
+ */
+(() => {
+  'use strict';
+
+  // Do nothing if `customElements` does not exist.
+  if (!window.customElements) return;
+
+  const NativeHTMLElement = window.HTMLElement;
```
notwaldorf: thinking out loud: we're using this on chrome and modern browsers so ES6 is fine, right?

webcomponents-ce-es5.js
```diff
@@ -0,0 +1,267 @@
...
+    nativeDefine.call(window.customElements, tagname, StandInElement);
+  };
+
+  window.customElements.get = (tagname) => constructorByTagname.get(tagname);
+
+})();
+
+/**
+ * @license
+ * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
```
notwaldorf: this file already has a license. also how can we have multiple self evaluating functions, is this a thing?

webcomponents-loader.js
```diff
@@ -9,8 +9,9 @@
  */
 
 (function() {
+  'use strict';
   // Feature detect which polyfill needs to be imported.
-  let polyfills = [];
+  var polyfills = [];
```
notwaldorf: thanks!

webcomponents-ce-es5.js
notwaldorf: Oh wait, am I reading an unminified bundle? Should we bundle?

webcomponents-ce-es5.js
azakus: Yep, and it also _must_ be in ES6 to function.

webcomponents-ce-es5.js
azakus: That's from the custom-elements code. Should I minify these bundles?

webcomponents-ce-es5.js
azakus: Same Q as https://github.com/webcomponents/webcomponentsjs/pull/711#discussion_r104508972

notwaldorf:
Uhhh I don't really understand anything in the es5 loader. there's much code, as if files got smooshed together. Is that a bundle? If so, should it be minified?

azakus:
@notwaldorf Yes, the ES5 bundle is just `rollup` on the HTMLImports polyfill and Custom Elements ES5 shim

webcomponents-es5-loader.js
```diff
@@ -0,0 +1,51 @@
...
+  // Feature detect which polyfill needs to be imported.
+  var polyfills = [];
+  if (!('import' in document.createElement('link'))) {
+    polyfills.push('hi');
+  }
+  if (!('attachShadow' in Element.prototype && 'getRootNode' in Element.prototype) ||
+    (window.ShadyDOM && window.ShadyDOM.force)) {
+    polyfills.push('sd');
+  }
+  if (!window.customElements || window.customElements.forcePolyfill) {
```
notwaldorf: This is the only difference from the regular loader, right?

azakus: Yes

bicknellr:
Does this PR mean I can remove the native shim from the custom-elements repo?

bicknellr:
Oh, never mind, it seems to import it from the custom elements repo in a few places. The reason I brought this up is that there's a copy of it being committed in this PR as part of webcomponents-ce-es5.js . Any way we can keep only one copy?

azakus:
@bicknellr That's not technically correct. The version in webcomponents is just a build artifact that uses the native-shim from custom-elements.

If you want to move the native-shim to webcomponents, that seems like another PR is in order.

ZepAviator:
Pulled the branch and did a `gulp default` build. 
Seems to step into another error when running on Chrome:
```
Uncaught TypeError: Cannot read property 'whenReady' of undefined
    at webcomponents-ce-es5.js:227
    at webcomponents-ce-es5.js:232
    at webcomponents-ce-es5.js:275
```

This is after compiling down `polymer build --js-compile`

bicknellr:
Ok, I didn't realize that this file was build output.

> If you want to move the native-shim to webcomponents, that seems like another PR is in order.

If the custom elements repo still contains the source as far as webcomponents.js is concerned, then there's not much reason to move it IMO.

azakus:
@ZepAviator Yep, found the bug, new build shortly.

README.md
```diff
@@ -61,6 +61,12 @@ elements, etc.). Here's an example:
 </script>
 ```
 
+## `webcomponents-es5-loader.js`
+
+Due to the requirement that Custom Elements be ES6 classes (https://html.spec.whatwg.org/multipage/scripting.html#custom-element-conformance), it may make sense to precompile your elements down to ES5 for wider consumption.
+However, this will conflict with the presence of native Custom Elements in browsers that support them.
```
justinfagnani: Custom Elements must be ES6 classes (https://html.spec.whatwg.org/multipage/scripting.html#custom-element-conformance). Since most projects need to support a wide range of browsers that don't necessary support ES6, it may make sense to compile your project to ES5. However, ES5-style custom element classes will not work with native Custom Elements because ES5-style classes cannot properly extend ES6 classes, like HTMLElement."

azakus:
@ZepAviator Actually, there is an issue right now with `polymer build` compiling the ES5 shim and/or not copying the bundles when using `webcomponents-loader`. So there's still some work to be done before `polymer build --js-compile` is there.

ZepAviator:
@azakus Is there a workaround for the `polymer build --js-compile` in the meantime?
I'm finding the same issues using my own gulp build task.

PR #725, opened 2017-03-12
Fix license value in bower.json

FluorescentHallucinogen:
According to bower.json spec (https://github.com/bower/spec/blob/master/json.md#license), "license" value must be one of SPDX license identifiers (https://spdx.org/licenses), "BSD" is not a valid SPDX identifier.


FluorescentHallucinogen:

PR #728, opened 2017-03-13
Fix description in package.json

FluorescentHallucinogen:


FluorescentHallucinogen:

notwaldorf:
Needs a rebase

FluorescentHallucinogen:
@notwaldorf
> Needs a rebase

Done.

FluorescentHallucinogen:
@valdrinkoshi, please merge this pull request.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "webcomponents.js",
   "version": "1.0.0-rc.8",
-  "description": "webcomponents.js",
+  "description": "Web components polyfill",
```
abdonrd: Perhaps this description would be better?

>A suite of polyfills supporting the Web Components specifications

FluorescentHallucinogen: 9 words vs 3. )))

abdonrd: Yes, but it is a description. haha :)

FluorescentHallucinogen: Do polyfills `supports` specifications? Polyfills _implement_ features (specification support) that not supported by browsers natively.

WebComponents.js is more than just a suite, it's a dynamic loader uses feature detection, ES5 adapter, etc...

Anyway, it's much better than original description — `webcomponents.js`.

Feel free to change the description after merge this pull request. But in this case don't forget to change it in `bower.json` file too.

FluorescentHallucinogen:
I believe we should use the same description in both `bower.json` and `package.json` files.

TimvdLippe:
Since we no longer have `bower.json`, I am closing this PR.

PR #730, opened 2017-03-13
[1/2] Update authors

FluorescentHallucinogen:


FluorescentHallucinogen:

notwaldorf:
This change isn't needed, since we're already linking to the authors file in all of the license headers ([example](https://github.com/webcomponents/webcomponentsjs/blob/master/src/pre-polyfill.js#L5)). Also, FYI, we use the "Polymer" not "Polymer Project" name.

PR #740, opened 2017-03-21
generated es5-loader

### LINKED ISSUE #722: Rename webcomponents-es5-loader to webcomponents-es5-transpiled-loader
The name is a bit confusing, as people _do not_ need to use the es5 loader unless they have transpiled code.

### LINKED ISSUE #737: Can't use ES5 with v1.0.0-rc.6
Hello,

So I'm trying out v1.0.0-rc.6 with the intent of supporting IE11 and better, but I'm facing a few issues:
- `webcomponents-lite.js` does not support ES5 code
- which leads me to try `webcomponents-es5-loader.js`, but it ends up [loading `webcomponents-lite.js` anyway](https://github.com/webcomponents/webcomponentsjs/blob/master/webcomponents-es5-loader.js#L34-L35).
- I tried loading `webcomponents-ce-es5.js` out of curiosity, but its use of arrow functions is not ES5 friendly...

Would it be possible to provide a version of `webcomponents-lite.js` with ES5 support?

Thank you.

valdrinkoshi:
Fixes #737  by generating `webcomponents-es5-loader.js` out of the custom elements native shim + webcomponents-loader.
This allows browsers that implement customElements (e.g. chrome, STP) to use es5 style js to define custom elements.
The gambit is: only the native-shim is needed synchronously when the script is executed. We bundle it into the loader for this reason.
This requires users to NOT compile `webcomponents-es5-loader.js` ⚠️ 

Here's an example of it working as expected http://jsbin.com/babifij/2/edit?html,console,output

Fixes #722 by renaming `webcomponents-loader` to `webcomponents-es6-loader`


azakus:
I like this approach. Let's also try #722 out in this PR.

valdrinkoshi:
@azakus ready for review! Firefox 52 timed out for some reason (can't repro locally), so I've kicked the build again https://travis-ci.org/webcomponents/webcomponentsjs/builds/213528757

azakus:
Closing in favor of #741

PR #741, opened 2017-03-21
Bundle custom-elements-es5-adapter

### LINKED ISSUE #737: Can't use ES5 with v1.0.0-rc.6

valdrinkoshi:
Fixes #737 by providing a new bundle `custom-elements-es5-adapter.js`, to be loaded by the user before declaring new Custom Elements

@azakus FYI


README.md
```diff
@@ -60,15 +60,19 @@ elements, etc.). Here's an example:
...
 
-To work around this, `webcomponents-es5-loader.js` first loads an extra ES5 compatibility [shim](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) before
-loading the minimum polyfill bundle, like `webcomponents-loader.js` does. 
+To work around this, load `custom-elements-es5-adapter.js` before declaring new Custom Elements. 
 
-The reason for having two different loaders (`webcomponents-loader.js` for ES6 deployed code and `webcomponents-es5-loader.js` for ES5) is that the polyfill can't know if you're deploying ES5 or ES6 code -- it runs before your code, and only _you_ as an author of the code know what you're going to deploy. This means that _you_ have to make the decision about which loader to use. 
-**TL; DR: ** Use `webcomponents-es5-loader.js` only if you want to deploy ES5 code to _all_ browsers (both with and without native custom elements support) -- if this isn't a requirement, then use
-`webcomponents-loader.js` instead.
+```html
+<script src="bower_components/webcomponentsjs/custom-elements-es5-adapter.js"></script>
```
notwaldorf: Don't you still need a loader for the actual polyfill?

valdrinkoshi: Yes you do, this example was for Chrome. I shall put the loader in here i guess..!

notwaldorf:
LGTM 2

MayhemYDG:
@valdrinkoshi 
Might be a good idea to compile the custom-elements-**es5**-adapter.js to **ES5**. Right now I'm seeing [arrow functions](https://github.com/webcomponents/webcomponentsjs/blob/f8b627e722e5d0d00fdb59c7753aa6f20aae739f/custom-elements-es5-adapter.js#L4).

valdrinkoshi:
@MayhemYDG the adapter **MUST** stay ES6 mainly because it has to extend `HTMLElement` class which doesn't have a constructor but can just be extended. See more details here https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js#L12

From my tests, in IE11/Edge loading `custom-elements-es5-adapter.js` doesn't cause any syntax error to be fired.

MayhemYDG:
Fair enough.

I do see a syntax error on IE11 though. :thinking: 
image

valdrinkoshi:
Oh weird. I've tested with the VM provided in modern.ie, with Windows 8.1 and IE11.
I would in fact expect it to throw a syntax error. Is there a way you can catch that error, e.g. onerror="" ?

007lva:
@valdrinkoshi I've tested with a VM with Windows 7 and IE11 and is clear that it doesn't support arrow functions.

robdodson:
My (rough) understanding is you'll get a syntax error in IE11 but it's harmless so long as the adapter is not bundled with the rest of your code.

robdodson:
actually i may be wrong, cc @justinfagnani to double check me on that :D

PR #745, opened 2017-03-22
Fixed issue with "window.ShadyCSS" not being defined for firefox 52

dfantal:
On firefox, "window.ShadyCSS" is not defined, so the value checks on values such as "window.ShadyCSS.ApplyShim" are undefined. Adding checks that "window.ShadyCSS" is defined to these checks fixes issues with firefox not loading correctly.


dfantal:
At the time I pushed this branch initially, the tests were failing on master. If you can trigger a rerun I bet they would pass now.

Also, it looks like the source files have been minified. Was that intentional. Seems like the source files should be somewhere else for these changes if they have been moved.

lukasoppermann:
Hey @dfantal, only you (by pushing an update) or someone with access to Travis can trigger a rerun.

ludwigschubert:
This issue is already resolved in ShadyCSS:
https://github.com/webcomponents/shadycss/blob/master/src/scoping-shim.js#L158

It looks like this PR directly modified a build product. (@dfantal let me know if I'm wrong, of course!) As far as I can see the current files on this repo already have this fix included--I was able to fix the issue when I encountered it by switching to the version of the polyfills in this repo.

I'd recommend to close this PR.

TimvdLippe:
This was fixed in https://github.com/webcomponents/shadycss/commit/23d877006539131343fe20225a2a696f625f3b4c

Issue #749, opened 2017-03-26
custom-elements-es5-adapter.js hilariously uses es6 code


jonmann20:
Somewhat ironically https://github.com/webcomponents/webcomponentsjs/blob/master/custom-elements-es5-adapter.js is in es6, so when trying to use it to polyfill IE 11, you get an error.

MartinsThiago:
This is expected to happen, es6 adapter target is to prevent errors is browser that support the v1 spec... because you cannot convert your elements to es5 and run on latest browsers

It's not so comfortable having error being thrown, but anyway... it works :)

rictic:
Yep, as @MartinsThiago says, this is necessary when shipping ES5 to a browser that supports native custom elements, as native elements are ES6 classes and when extending an ES6 class you _must_ call `super()`, and there's no ES5 way to do that.

The syntax error being thrown in IE11 and friends is an issue though, and we're tackling it in `polymer build` here: https://github.com/Polymer/polymer-cli/issues/627

MartinsThiago:
Maybe we should consider renaming this file
es5-adapter... may lead people to think it's a polyfill to older browsers

_Naming suggestion: custom-elements-run-your-es5-compiled-element-on-modern-browsers.js_

jonmann20:
Ah I see, ok thank you for clarifying!

CITguy:
I'm still confused. How do I use the adapter if it's meant for ES5 browsers that don't support ES6, but it's written with ES6 syntax?

valdrinkoshi:
The ES5 adapter is needed for native Custom Elements.
Native Custom Elements must be defined with a class (ES6), but most projects transpile to ES5 for reasons (e.g. better tooling available for minification/bundling, or support older browsers).
When you transpile to ES5, you need an adapter for native Custom Elements to work (e.g. in Chrome or Safari).
When you load the adapter in a browser that doesn't support ES6, it will fire a syntax error, and that's expected to happen, because you cannot feature detect the support of ES6 at run time.

In short: the ES5 adapter is written in ES6 to enable native Custom Elements to be declared in ES5 syntax

rictic:
I think we should rename the shim one last time, as this is a repeated point of confusion.

Serious proposal: `es6-shim-for-es5-in-ce-browsers.js`

jehna:
How about `customelements-fix-es5-new-htmlelement-error-in-ce-browsers.js`?

svastanesto:
[...]

PR #757, opened 2017-04-06
Clarify purpose of custom-elements-es5-adapter.js

robertwbradford:
There is some confusion as to the purpose of the ES5 adapter (e.g., https://github.com/webcomponents/webcomponentsjs/issues/749) This clarifies that the adapter is not an ES5 polyfill but rather that it should be used if a project has been compiled to ES5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ elements, etc.). Here's an example:
 ```
 
 ## `custom-elements-es5-adapter.js`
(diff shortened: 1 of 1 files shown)


TimvdLippe:
@robertwbradford Sorry for the long wait. This seems like a good change to me. Do you mind rebasing?

TimvdLippe:
Merged manually, thanks!

PR #762, opened 2017-04-18
Fix `webcomponents-loader` on older browsers that support native HTML imports

sorvell:
* Load full suite of polyfills (`-lite`) if ES6 features (e.g. Array.from) are not available. We do this because if ES6 is not available, none of the other webcomponents features (which require ES6) are definitely not available.
* Uses `document.write` to load polyfills if and only if native HTML imports exists (older Chrome browsers). This ensures that polyfills are loaded by the time scripts in HTML imports run.
* Adds old Chrome version to test matrix.


webcomponents-loader.js
```diff
@@ -42,7 +40,14 @@
     var replacement = 'webcomponents-' + polyfills.join('-') + '.js';
     var url = script.src.replace(name, replacement);
     newScript.src = url;
-    document.head.appendChild(newScript);
+    // NOTE: this is not CSP compliant but it's required to ensure the polyfills
+    // are loaded before *native* html imports load on older Chrome versions.
+    // In all other cases, this can be async.
+    if (document.readyState === 'loading' && ('import' in document.createElement('link'))) {
+      document.write(newScript.outerHTML);
```
notwaldorf: :sob:

notwaldorf: (i read the comment, i understand, i can still sob)

valdrinkoshi:
Chrome 41 will need also customElements, that's the main issue (thanks @azakus for explaining that offline). But maybe that could be stubbed too..!

sorvell:
Unfortunately, we can't do an early patch like that. If the custom elements polyfill is in use then `HTMLElement` must be patched when a class is defined, not when `customElements.define` is called.

kevinpschaaf:
The gambit to stub `customElements.define` in the loader hit a rough patch, because even when stubbed the patched `HTMLElement` does not yet exist, so classes that extend `HTMLElement` will not be usable, so at the limit the "stub some minimal stuff" starts to approach moving a lot of chunks of the polyfill into the loader, which defeats the purpose of the loader and makes it way harder to maintain.

The good news is we did some de-FUD'ding, and `document.write('<script src="..."></script>')` does _not_ in fact violate CSP any more than loading a script normally via a static `<script src="..."></script>` would.  In both cases the domain the script will be loaded from must be whitelisted (e.g. `script-src 'self'`), but is otherwise totally groovy with CSP.  See [`document.write()` treatment here](http://www.cspplayground.com/violation_examples) and [demo of this working under CSP here](http://localhost/apache/dev/polymer/projects/web-test/csp-write.html).

Last, the change made in [Chrome 55 to disable cross-domain document.write on 2G connections](https://developers.google.com/web/updates/2016/08/removing-document-write) would never hit this case, since the `document.write` introduced here would only run on Chrome 33..53 (versions _with_ HTML imports but _without_ Custom Elements V1).

As such, this approach seems sound.

webcomponents-loader.js
```diff
@@ -23,26 +23,30 @@
...
     !(document.createDocumentFragment().cloneNode() instanceof DocumentFragment)) {
-    polyfills.push('pf');
-  }
-
-  if (polyfills.length === 4) { // hi-ce-sd-pf is actually called lite.
     polyfills = ['lite'];
   }
 
   // NOTE: we stub HTMLImports to enable WCT to wait for "WebComponentsReady" event.
-  window['HTMLImports'] = {};
```
valdrinkoshi: This will be run also when we force the polyfills, so also in Chrome >= 55 :/
I don't know how worth it is, but what if we document.write an import instead of a script here?

valdrinkoshi:
@kevinpschaaf `document.write` will be executed also in Chrome 55 or superior if we force the polyfills :/
We might be able to workaround the skipping of the scripts if we `document.write` an import instead of a script maybe?

valdrinkoshi:
Tested the `document.write` of an import in Chrome 57 and I don't get any console warning 😁 🎉  http://jsbin.com/wapacad/6/

webcomponents-loader.js
kevinpschaaf: We were discussing changing [this line](https://github.com/Polymer/web-component-tester/blob/master/browser.js#L61) in WCT to `if (!window.HTMLImports && !document.querySelector('script[src*="webcomponents-"]')` (basically, loading any variant of `webcomponents-*.js` should guarantee that `WebComponentsReady` is fired.

@sorvell Was the intention to go forward and propose that change along with removal of the `HTMLImports` stub object above?

valdrinkoshi: Done in https://github.com/Polymer/web-component-tester/pull/533

PR #791, opened 2017-05-26
Fix build instructions

fredj:
There's no 'build' target in the gulpfile.


Haprog:
I just noticed the same thing, and was going to make a PR until I noticed this.

justinfagnani:
Thanks @fredj !

PR #793, opened 2017-05-27
Add a file with compiler externs for the API exposed by the polyfills.

rictic:
/cc @azakus

### DIFF
--- a/externs/api.js
+++ b/externs/api.js
@@ -1,0 +1,29 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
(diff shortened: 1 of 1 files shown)


TimvdLippe:
@rictic is this still something you would like to pursue?

rictic:
I think this is desirable, yeah. I don't have rights to assign a reviewer, but I think Dan would be a good one for this.

azakus:
Would it be better for this to go into HTMLImports?

valdrinkoshi:
FYI there is already one in html-imports https://github.com/webcomponents/html-imports/blob/master/externs/html-imports.js

azakus:
@valdrinkoshi maybe port this one over there because it has comments 😁

valdrinkoshi:
Done https://github.com/webcomponents/html-imports/pull/80

rictic:
Closed in favor of webcomponents/html-imports#80

Issue #801, opened 2017-06-19
allow using the webcomponents-loader.js script inline


vlukashov:
When optimizing page load performance, as a developer I may want to reduce the number of HTTP round trips required to render a page. The webcomponents-loader script is small and therefore looks like a perfect candidate for inlining. However, as it is currently written, it does not work when inlined (since it tries to find the `<script>` tag that includes itself, and use its `src` url).

Please add an option to support inlining webcomponents-loader.js.

One approach to that could be  that the base URL for dynamically loaded polyfills is set through the global Web Components settings:

```html
<script>
  window.WebComponents = window.WebComponents || {};
  window.WebComponents.base = 'bower_components/webcomponentsjs/';
</script>
<script>
  /* webcomponents-loader.js (with the modifications to check WebComponents.base) */
</script>
```

trentmwillis:
I've opened a PR to fix this at #802.

Haprog:
We're currently using a custom fork of `webcomponentsjs` just because of this, so it would be nice to have the fix merged in, so that we wouldn't have to maintain the fork when `webcomponentsjs` gets updates.

vlukashov:
Google PageSpeed has issues with the webcomponents-loader.js likely due to this same issue (see #806).

A humble reminder to the maintainers (and that seems to be @sorvell currently):

Could you please indicate--as the maintainers of the repo--if you see the ability to inline the loader a useful feature at all? It would improve the compatibility with Google PageSpeed (and will also allow for performance improvements) at the cost of a few extra lines of code.

And if this feature is useful, could you please review the PR and list the missing bits that block it from being merged? (I assume documentation update would be one). I believe the community could help you out implementing this feature if there is a clear understanding of what needs to be done.

PR #802, opened 2017-06-20
Support inlining webcomponents-loader via WebComponents.root property

trentmwillis:
Fixes https://github.com/webcomponents/webcomponentsjs/issues/801.

If a user inlines `webcomponents-loader.js` (via the Polymer-CLI for instance), then this will allow the loader to continue working so long as they specify a value for `window.WebComponents.root`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,15 @@ Here's an example:
 </script>
 ```
 
+If you have inlined the source of `webcomponent-loader.js`, then you should specify `window.WebComponents.root` as the root from which to load the polyfills. For example:
+
+```html
+<script>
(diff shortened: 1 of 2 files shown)


trentmwillis:
Any chance this could get looked at? Currently, this issue prevents us from optimizing our Polymer app because we can't inline this script for non-Chrome browsers.

Haprog:
Please move this forward 🙂

trentmwillis:
Thanks for merging! Is there any chance this could get back-ported to the 1.x release branch? We have apps that still use HTML Imports and being able to use this with those apps would be hugely beneficial.

PR #844, opened 2017-09-11
Fix Promise.prototype.catch closure issue

### LINKED ISSUE #837: regression: Promise.prototype.catch is undefined in IE11 (v1.0.8+)
Starting from the version 1.0.8 of the WebComponents polyfill, the ES6 promise polyfill shipped as a part of webcomponents-lite.js does not define the `Promise.prototype.catch` method. For example, the [app-storage](https://www.webcomponents.org/element/PolymerElements/app-storage) web components break the app in IE11 starting with WebComponents polyfill 1.0.8 (1.0.7 is still OK).

Here is a demo app (broken in IE11): https://github.com/vlukashov/polymer-local-storage.

Most likely, the root cause is [closure-compiler/#2640](https://github.com/google/closure-compiler/issues/2640), but for some reason the issue did not surface with the polyfill versions 1.0.7 and before.

**Workaround**
Pin the version of the WebComponents polyfill to 1.0.7 or less.

azakus:
Manually import ES6Promise polyfill and closure export `catch` function
Fixes #837


src/promise.js
```diff
@@ -0,0 +1,20 @@
...
+'use strict';
+import ES6Promise from '../node_modules/es6-promise/lib/es6-promise.js';
+
+/*
+Assign the ES6 promise polyfill to window ourselves instead of using the "auto" polyfill
+to work around https://github.com/webcomponents/webcomponentsjs/issues/837
+*/
+if (!window.Promise) {
+  window.Promise = ES6Promise;
+  ES6Promise.prototype['catch'] = ES6Promise.prototype.catch;
```
keanulee: :bleeding_eyes:

azakus: Oh sweet summer child...

notwaldorf: popcorn.gif

andreabioli:
I know it's a stupid question, but how to get this fix in my local application, please?

keanulee:
You can update `bower.json` dependencies to include `"webcomponentsjs": "master"` or wait until a new release is published.

@azakus good time to tag a new release?

PR #851, opened 2017-09-23
Remove generating blank .map file for webcomponents-lite.js

paladox:
It was failing to load correctly due to a blank .map file.

As it's blank lets remove it for webcomponents-lite.js.


TimvdLippe:
This was superseded and fixed by #858

paladox:
thanks

PR #858, opened 2017-10-13
fix the issue with non-inspectable polyfills in dev tools

### LINKED ISSUE #812: Include sourcesContent into source maps (regression from v1.0.1)
Starting from the version 1.0.2 webcomponentsjs source maps do not include the `sourcesContent` section (it's empty) and therefore in order to actually use the source maps, one needs to host the original source files along with the webcomponentsjs polyfills.

webcomponents-lite.js.map version history:
 * [1.0.1](https://raw.githubusercontent.com/webcomponents/webcomponentsjs/e3558c3b9128f27c4f41c54f4fb59553e23748ce/webcomponents-lite.js.map) `sourcesContent` is OK
 * [1.0.2](https://raw.githubusercontent.com/webcomponents/webcomponentsjs/bb644a6f250f57b49b336370949edc6c18c9af4c/webcomponents-lite.js.map) `sourcesContent` is empty
 * [1.0.3](https://raw.githubusercontent.com/webcomponents/webcomponentsjs/fb6034def21f893e81b43e23329295556bd0951b/webcomponents-lite.js.map) `sourcesContent` is empty

Please update the builds so the full set of the original source files is again included into the source maps files.

vlukashov:
Due to a misconfiguration of the gulp build pipeline, the source maps for all minified polyfill files are generated without including the contents of the original source files. This causes the polyfills to be non-inspectable in the browser developer tools (compare the before and after screenshots below).

Before:
before-sources-are-not-inspectable

After:
after-sources-are-inspectable

Fixes #812

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -75,7 +75,7 @@ function closurify(sourceName, fileName) {
   };
 
   return rollup(rollupOptions)
-  .pipe(source(`${sourceName}-index.js`, 'entrypoints'))
+  .pipe(source(`${sourceName}-bundle.js`))
   .pipe(buffer())
   .pipe(sourcemaps.init({loadMaps: true}))
   .pipe(closure(closureOptions))


TimvdLippe:
Oh cool! Yes I hit this issue as well and didnt know what the fix would be. Would be great to see this fixed :+1:

limonte:
a41

gulpfile.js
```diff
@@ -75,7 +75,7 @@ function closurify(sourceName, fileName) {
   };
 
   return rollup(rollupOptions)
-  .pipe(source(`${sourceName}-index.js`, 'entrypoints'))
+  .pipe(source(`${sourceName}-bundle.js`))
```
azakus: Can you explain why this change works? Nothing seems obvious.

vlukashov: The build pipeline here has two transformation steps: **rollup** and **closure**, and the **sourcemaps** plugin that hooks in between to ensure that in the end the source map correctly links the final output to the original source files.

The output of the **rollup** step is not persisted into a file on disk, yet the rest of the pipeline expects to get something looking like a regular file (with a file name and path). That's what the ``source(`${sourceName}-index.js`, 'entrypoints')`` line is doing: it makes the **sourcemaps** and the **closure** steps down the pipeline believe they work with `entrypoints/${sourceName}-index.js`.

**rollup** is configured to create a source map (`sourceMap: true,`). It does so and adds it into the output as an inline base64-encoded URL. That source map references the original source files in 'entrypoints', 'src' and 'node_modules' with absolute paths. In the `webcomponents-hi-index.js` case it looks like this:
```javascript
{
    version: 3,
    file: null,
    sources: [
        "/Users/viktor/projects/webcomponents/webcomponentsjs/node_modules/@webcomponents/html-imports/src/html-imports.js",
        "/Users/viktor/projects/webcomponents/webcomponentsjs/src/post-polyfill.js",
        ...
    ],
    ...
}
```

**sourcemaps** reads that inline source map data (`sourcemaps.init({loadMaps: true})`) and attaches it to the `entrypoints/${sourceName}-index.js` file stream that gets passed into **closure**. When doing so **sourcemaps** converts the paths to the original source files into the relative form. In case of `webcomponents-hi-index.js` it looks like this:
```json
[
  "../node_modules/@webcomponents/html-imports/src/html-imports.js",
  "../src/post-polyfill.js",
  ...
]
```
Note 2 things about these source file references:
 - the paths here are relative as if the file referencing them was in the `entrypoints` directory
 - there is a reference to `webcomponents-hi-index.js` which looks like the file is referencing itself as its source

Both of these are the results of the ``source(`${sourceName}-index.js`, 'entrypoints')`` line.

The **closure** plugin at the next step converts the `entrypoints/${sourceName}-index.js` file stream coming out of **sourcemaps** (alone with the source map data and source file paths) into JSON and passes that into google closure compiler (jscomp). This is where the `entrypoints` part of `entrypoints/${sourceName}-index.js` disappears - jscomp gets only the filename (without the preceding directory), yet the paths in the accompanying source map remain unchanged. This looks like a bug in the **closure** gulp plugin, but I cannot tell.

Anyway, for the `webcomponents-hi-index.js` case the JSON data sent to jscomp looks like this:
```javascript
{
    src: "...",
    path: "webcomponents-hi-index.js",
    sourceMap: {
        version: 3
        file: "webcomponents-hi-index.js",
        sources: [
            "../node_modules/@webcomponents/html-imports/src/html-imports.js",
            "../src/post-polyfill.js",
            ...
        ],
        ...
    }
}
```
In response jscomp returns a JSON object including an updated source map, but the issue is still there. The relative paths to the source files do not match to the path of the output (the 'entrypoints' part is missing).

For the `webcomponents-hi-index.js` case the JSON data received back from jscomp looks like this:
```javascript
{
    src: "...",
    path: "webcomponents-hi.js",
    source_map: {
        version:3,
        file: "webcomponents-hi.js",
        sources: [
            "../node_modules/@webcomponents/html-imports/src/html-imports.js",
            "../src/post-polyfill.js",
            ...
        ],
        ...
    }
}
```
Note that jscomp kept the relative paths as they were.

Finally, when **sourcemaps** creates the final source map it adds the original sources contents to the source map object returned by jscomp, but since the relative paths there are wrong, it fails to do so and the `sourcesContent` property of resulting the source map file appears empty.

---
This PR fixes the issue by changing the name and location of the virtual file produced by **rollup** to be `${sourceName}-bundle.js`. That helps in two ways:
 - the contents of the original source files are correctly included into the final source maps (because there is no additional directory that can be dropped and mess the relative file paths)
 - references to the **rollup** output file can be clearly separated from the references to the original `${sourceName}-index.js` file (they would appear referencing `${sourceName}-rollup.js`)

azakus: Awesome, thanks for the very detailed information 👏

PR #861, opened 2017-10-26
Allow loader to run in body

CITguy:
* still works in `<head>`
* works within container anywhere in `<body>`, too 
  * must be wrapped in a container (`<span><script src="webcomponents-loader.js"></script></span>`)


CITguy:
I would like this to be available, so that I can organize all of my javascript at the bottom of the `<body>` rather than split between the `<head>` and `<body>`.

TimvdLippe:
This should now be working, tested in https://github.com/webcomponents/webcomponentsjs/blob/master/tests/loader-after-load.html

Issue #817, opened 2017-07-28
Firefox nightly needs webcomponentsjs/webcomponents-hi-sd.js


sdwlig:
Firefox nightly apparently now satisfies Custom Elements v1, so the loader gets a 404 looking for:
webcomponentsjs/webcomponents-hi-sd.js

justinfagnani:
We need to verify it's a working implementation, but... awesome news!

justinfagnani:
I'm seeing that `window.customElement` is undefined. Did you turn on a flag?

sdwlig:
I just found that I did turn it on in an earlier debugging session.  I'm not sure that it is working well enough, and have turned it off again for now.  Not sure whether webcomponentsjs can verify if it is all there or not.  In any case, generating a 404 is probably not the best response in this situation.  I copied webcomponents-hi-sd-ce.js to webcomponents-hi-sd.js, but this didn't seem to actually work.  Either it doesn't have a working custom elements implementation or the polyfill doesn't work with it enabled.  Our now-ported-to-Polymer 2.0 code works great in Chrome but is broken again in Firefox / Firefox nightly.  Earlier in our port, I found that I had a sequence assumption, now fixed.

bennypowers:
Nightly now ships with `dom.webcomponents.customelements.enabled` which is causing breakage along these lines.

web-padawan:
See also [#874(comment)](https://github.com/webcomponents/webcomponentsjs/issues/874#issuecomment-357109921)

azakus:
After some testing, Firefox Nightly is less than ready to go for Custom Elements. Upgrades seem to be in a funky order due to https://bugzilla.mozilla.org/show_bug.cgi?id=1326028. Elements that were parsed in the main page may upgrade before elements created dynamically, instead of a strict document ordering as required in the spec.

bennypowers:
FYI The bug @azakus linked to is resolved and customElements have landed in 59 stable. I was able to run a simple app composed from LitElements by enabling the `customElements`, `shadowDom`, and `moduleScripts` flags in plain old Firefox 59, with just the HTML imports polyfill for a few simple polymer2 elements. I wasn't able to get ShadyCSS working though, so polymer1/2-era elements that use `@apply` did not render properly.

Try it yourself: https://mws.bennypowers.com

PR #886, opened 2018-01-13
Redirect webcomponents-hi-sd.js to webcomponents-hi-sd-ce.js

MajorBreakfast:
Interim solution for #817 because anything is better that having it not work at all. Currently `webcomponents-hi-sd.js` does not exist, this consequently leads to a 404. This PR just fixes this by loading `webcomponents-hi-sd-ce.js` even though the custom elements polyfill isn't need.


TimvdLippe:
Closing per https://github.com/webcomponents/webcomponentsjs/issues/817#issuecomment-359529687

Issue #305, opened 2015-05-15
cca (Chrome App Mobile) error in hello world test 


jmorille:
I have written a hello world test in cca v0.7.0
with Polymer/polymer#^0.5.6 and "webcomponentsjs": "0.7.0", (I am tested also the webcomponentsjs:0.5.5)

 An error 

```console
GET chrome-extension://jjpffmggdecnnmfpffjkmcgmcclabdce/plugins/org.chromium.bootstrap/chromeappstyles.css 404 (Not Found) window.js:203
'window.webkitStorageInfo' is deprecated. Please use 'navigator.webkitTemporaryStorage' or 'navigator.webkitPersistentStorage' instead. webcomponents.js:210
Uncaught NotFoundError: Failed to execute 'appendChild' on 'Node': The new child element is null. webcomponents.js:4633
```

that correspond of the corresponding code of the webcomponents.js:4633
```javascript
  function addCssToDocument(cssText) {
      if (cssText) {
        getSheet().appendChild(document.createTextNode(cssText));
      }
    }
```

The source code to reproduce this error is very simple, It just need to initialize the dependencies
```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">

    <link rel="manifest" href="manifest.json">

    <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
    <meta name="theme-color" content="#303F9F">
    <!-- Enable "installability" -->
    <meta name="mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-capable" content="yes">
    <meta name="apple-touch-fullscreen" content="yes">
    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" >
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />

    <!-- build:css styles/main.css -->
    <link rel="stylesheet" href="styles/main.css" shim-shadowdom>
    <!-- endbuild-->

    <!-- build:js scripts/vendor/webcomponents.js -->
    <script src="bower_components/webcomponentsjs/webcomponents.js"></script>
    <!-- endbuild -->


    <link rel="import" href="bower_components/polymer/polymer.html">

 

<body>

<h1>Hello</h1>


</body>
</html>
```

Before an 404 error occur about loading chromeappstyles.css 
[...]

PR #903, opened 2018-03-15
Upgrade es6-promise

### LINKED ISSUE #794: vertx dependency in webcomponents-lite.js
What is this `vertx` and why is it required in try-catch within `webcomponents-lite.js` ? there is nothing like that in dependencies of this project

https://github.com/webcomponents/webcomponentsjs/blob/master/webcomponents-lite.js#L94

mroem:
- Update `es6-promise` to `4.2.4` to fix [unresolved `vertx`](https://github.com/stefanpenner/es6-promise/issues/305)

Fixes #794 

*UPDATE*
- Reverting upgrade
- Use `patch-package` to resolve issue using related fix: see [#305 on `es6-promise`](https://github.com/stefanpenner/es6-promise/pull/323)


TimvdLippe:
Hm, although I just inspected the Travis build log and it does seem related to `Promise`. Maybe it actually broke with this upgrade 😢

mroem:
*UPDATE*

- Use patch-package to resolve issue using related fix: see #305 on es6-promise

TimvdLippe:
I would rather not have a private patch applied, as that is brittle going forward with later versions of the polyfill. Rather, we should investigate why the build started failing and report issues to the author to get them resolved.

mroem:
_UPDATE_

The es6-promise package needs to be transpiled as it uses Classes which causes `google-closure-compiler` to nullify some properties in ADVANCED mode.
See `Static class functions can be erroneously removed in ADVANCED mode` [#2763](https://github.com/google/closure-compiler/issues/2763)

TimvdLippe:
@mroem I have changed the base to v2. Could you resolve the conflicts and try again? v2 should be working on Travis (barring some test failures).

I do want to get this PR in, but I think we have to wait till the v2 work has been done before making this change. Thank you for your patience and development on this!

mroem:
@TimvdLippe 
How much work is still needed on v2?
I kind of need this change urgently, and it would solve this issue for all the people using v1.
I‘m happy to put a second Pull request up for v2, so that we can merge this into v1

NMinhNguyen:
@TimvdLippe could you elaborate on the reasons for having to wait till the v2 work has been done?

TimvdLippe:
It will be released next week. New additions should be added there, to
prevent diverging the two versions too much.

NMinhNguyen:
> to prevent diverging the two versions too much.

@TimvdLippe correct me if I’m misunderstanding, but if this change was applied to both branches, there’d be no divergence, right? Good to hear v2 is coming soon though 🙂

mroem:
@TimvdLippe @azakus @NMinhNguyen any update on this?

TimvdLippe:
Test failures are the same on this branch and on master, so this change seems fine to me. Thanks @mroem for keeping this PR up-to-date!

PR #907, opened 2018-03-26
Module compat loader

### LINKED ISSUE #870: Polyfills without HTML Import
I'm not using HTML imports in my custom element library (I build a single JS asset for my custom elements). However, I have no way of excluding the HTML Import polyfill from the loader.

What should I do to get a custom build to exclude HTML import polyfills?

### LINKED ISSUE #880: WC loader for JS modules
The move from HTML Imports to JS Modules requires changes to the webcomponents loader. Namely, we now have browsers that support Modules but not Custom Elements/Shadow DOM. This issue discusses approaches to a single “universal” build that optimally loads only the required polyfills on all browsers. It assumes you have no “differential serving” to specific user agents (in that case you can just include the required polyfills for that browser).

### Dynamic import(), JS Modules, ES6, and CE/SD
- Chrome Canary with experimental platform features flag

These browsers need no polyfills and can directly use the output of (in-development) polymer-bundler.

### JS Modules, ES6, and CE/SD
- Chrome
- Safari

These browsers need no polyfills but require a dynamic loader (such as Webpack).

### JS Modules, ES6, but no CE/SD
- Edge

These browsers require just webcomponents-sd-ce.js.

The existence of this combination means we cannot rely on <script nomodule> to selectively load polyfills. Also, attempting to load webcomponents-sd-ce.js as a module in Edge results in an “Unable to get property 'Object' of undefined or null reference” error.

### ES6, but no JS Modules nor CE/SD
- Firefox

These browsers require just webcomponents-sd-ce.js.

### Nothing
- IE 11

These browsers require CE/SD polyfills, transpilation to ES5, as well as some other polyfills (<template>, Promise, Array.from, fetch()).

The existence of this combination means we need a CE “class” adapter that selectively loads on native CE browsers. (At least currently,) if a browser supports native CE, then it also supports modules, so we can use <script type=”module”> to do this.

## Loading polyfill code
We only want to load the minimal polyfill code required and optimize for browsers that natively support APIs. One approach is to wrap entrypoint code with feature detection:

```js
function main() {
  import(/* webpackMode: 'eager' */ './components/hn-app.js');
}

if (!window.customElements || !('attachShadow' in Element.prototype)) {
  if (window.fetch) {
    // Firefox/Edge
    import(/* webpackChunkName: 'sd-ce-polyfill' */ '../node_modules/@webcomponents/webcomponentsjs/webcomponents-sd-ce.js').then(main);
  } else {
    // IE
    // No native Promise, so can't use Webpack's import();
    var script = document.createElement('script');
    script.src = '/polyfills/webcomponents-lite.js';

    var fetchScript = document.createElement('script');
    fetchScript.src = '/polyfills/fetch.js';

    fetchScript.onload = script.onload = () => {
      if (window.customElements && window.fetch) main();
    };

    document.head.appendChild(script);
    document.head.appendChild(fetchScript);
  }
} else {
  main();
}
```

Notes:
- `/* webpackMode: 'eager' */` tells Webpack to include the code of the entrypoint in this chunk so there’s no additional network request (Webpack docs)
- My sample app requires fetch() specifically, but one could imagine other polyfills here
  - Maybe bundle into a custom-built webcomponents-lite.js?
- HTML Imports polyfill should be removed from webcomponents-lite.js
  - Perhaps renamed as well

## Sample App
Demo: https://polymer-redux-hn.appspot.com/

Code: https://github.com/keanulee/polymer-redux-hn/tree/es5-webpack

azakus:
1. Removes HTML Imports (and previous CE flush polyfill callback coordination between CE & HI polyfill)
2. Renames `webcoimponents-lite.js` to `webcomponents-bundle.js` (we need to discuss the cost/benefit of this given all tests everywhere need to update)
3. Adds a `WebComponents.whenReady(cb)` method in `webcomponents-loader.js` that will be the primary way to do "easy mode polyfill loading" with modules; the callback can return a promise, and CE polyfill flushing will be held off until the promise resolves (to give rough equivalency with the performance optimization we had in HTML Imports).  

So the pattern for loading app code from `index.html` would be like this:

```html
<script src="webcomponents-loader.js" defer></script>
<script type="module">
  WebComponents.whenLoaded(function() {
    return import('app.js');
  });
</script>
```

Closes #870 
Closes #880


webcomponents-loader.js
```diff
@@ -31,44 +121,25 @@
   if (!('content' in document.createElement('template')) || !window.Promise || !Array.from ||
```
ExE-Boss: The full suite has been renamed to `sd-ce-pf` in this PR.

entrypoints/webcomponents-bundle-index.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function finish() {
+  requestAnimationFrame(() => {
+    window.WebComponents.ready = true;
+    document.dispatchEvent(new CustomEvent('WebComponentsReady', { bubbles: true }));
+  });
+}
+
+if (document.readyState === 'loading') {
```
kevinpschaaf: Remove the `rAF` above and if not `'complete'` wait for load event

webcomponents-loader.js
```diff
@@ -31,44 +121,25 @@
...
-    // *native* html imports load on older Chrome versions. This *is* CSP
-    // compliant since CSP rules must have allowed this script to run.
-    // In all other cases, this can be async.
-    if (document.readyState === 'loading' && ('import' in document.createElement('link'))) {
-      document.write(newScript.outerHTML);
-    } else {
-      document.head.appendChild(newScript);
-    }
+    newScript.addEventListener('load', function() {
+      ready();
```
sorvell: It seems like an unnecessary change from the previous behavior to divorce the `WebComponentsReady` event from `whenLoaded`. Specifically, previously `WebComponentsReady` used to mean "all imports that likely include the custom elements" have loaded. Now it just means "the polyfills have loaded and maybe some other stuff has."

Another related issue, it seems like if you never call `whenLoaded`, custom elements will never upgrade.

Perhaps instead we can:

1. fire the `WebComponentsReady` event after the Promise returned from `whenLoaded` resolves one is returned or when it is called. 

2. If `whenLoaded` has not been called by the time the document has loaded, call it.

This or some flavor of it would solve both issues.

azakus: `1` is already what is happening.

Custom Elements will absolutely upgrade without calling `whenLoaded`. The only way they would never upgrade is if `whenLoaded` is called, and the function or promise given there never resolves.

sorvell: Or just remove the event if we can.

entrypoints/webcomponents-bundle-index.js
```diff
@@ -0,0 +1,49 @@
...
+import './webcomponents-sd-ce-pf-index.js';
+
+const customElements = window.customElements;
+const Template = window.HTMLTemplateElement;
+
+if (customElements['polyfillWrapFlushCallback'] && Template.bootstrap) {
+  let shouldFlush = false;
+  /** @type {?function()} */
+  let flusher = null;
+  customElements['polyfillWrapFlushCallback']((flush) => {
```
valdrinkoshi: I got a bit confused trying to follow this code, as it seems it might execute `flush` twice.
Would it be less confusing to do something like this?
```js
customElements['polyfillWrapFlushCallback']((flush) => {
  if (shouldFlush) {
    flush();
  } else {
    flusher = flush;
  }
}));

window.addEventListener('DOMContentLoaded', () => {
  if (flusher) {
    flusher();
  } else {
    shouldFlush = true;
  }
});

**Comment by ExE-Boss** [comment]
The `webcomponents-loader.js` file still refers to `webcomponents-lite.js`, despite the fact that this PR renames it to `webcomponents-sd-ce-pf.js`: https://github.com/webcomponents/webcomponentsjs/blob/86bbf629ad774ff8fa1164f94cec35be62d3a8d2/webcomponents-loader.js#L85-L91

File: README.md [hunk 5]
```diff
@@ -1,98 +1,94 @@
...
+### Using `webcomponents-loader.js`
 
-Additionally, you can check if `window.WebComponents` exists to know if the `WebComponentsReady` event will fire, and you can check if `window.WebComponents.ready` is true to check if the `WebComponentsReady` event has already fired.
+The `webcomponents-loader.js` is a client-side loader that dynamically loads the
+minimum polyfill bundle, using feature detection. Note that because the bundle will be
+loaded asynchronously, scripts that depend on webcomponents APIs *must* be loaded
+using `WebComponents.waitFor(loadCallback)`, for example custom elements definitions. The `loadCallback` function should load scripts that need the polyfills (typically via `import('my-script.js')`) and
+should return a promise that resolves when all scripts have loaded.
 
 Here's an example:
```
sorvell: Let's add an example that doesn't use `waitFor`

README.md
```diff
@@ -127,37 +122,27 @@ Everything in this repository is BSD style license unless otherwise specified.
...
-```js
-window.addEventListener('WebComponentsReady', function(e) {
-  // imports are loaded and elements have been registered
-  console.log('Components are ready');
-});
-```
+* The HTML Imports polyfill has been removed. Given that ES modules have shipped in
+most browsers, the expectation is that web components code will be loaded via
+ES modules.
+* When using `webcomponents-loader.js`, scripts that rely on the polyfills *must* be loaded using `WebComponents.waitFor(loadCallback)`.
```
sorvell: Add node that this is only requited if `defer` is used on the script.

README.md
```diff
@@ -1,98 +1,94 @@
...
+- **Custom Elements v1**: allows authors to define their own custom tags ([spec](https://w3c.github.io/webcomponents/spec/custom/), [tutorial](https://developers.google.com/web/fundamentals/getting-started/primers/customelements), [polyfill](https://github.com/webcomponents/custom-elements)).
+- **Shadow DOM v1**: provides encapsulation by hiding DOM subtrees under shadow roots ([spec](https://w3c.github.io/webcomponents/spec/shadow/), [tutorial](https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom),
+[shadydom polyfill](https://github.com/webcomponents/shadydom), [shadycss polyfill](https://github.com/webcomponents/shadycss)).
 
 For browsers that need it, there are also some minor polyfills included:
 - [`HTMLTemplateElement`](https://github.com/webcomponents/template)
 - [`Promise`](https://github.com/stefanpenner/es6-promise)
-- `Event`, `CustomEvent`, `MouseEvent` constructors and `Object.assign`, `Array.from` (see [webcomponents-platform](https://github.com/webcomponents/webcomponents-platform))
+- `Event`, `CustomEvent`, `MouseEvent` constructors and `Object.assign`, `Array.from`
```
sorvell: add URL polyfill here.

entrypoints/webcomponents-bundle-index.js
```diff
@@ -0,0 +1,40 @@
...
+import './webcomponents-sd-ce-pf-index.js';
+
+const customElements = window.customElements;
+const Template = window.HTMLTemplateElement;
+
+let shouldFlush = false;
+/** @type {?function()} */
+let flusher = null;
+
+if (customElements['polyfillWrapFlushCallback'] && Template.bootstrap) {
```
sorvell: Remove check for Template.bootstrap since we always want to do this if we're polyfilling custom elements.

tests/ce-import-upgrade-async.html
```diff
@@ -13,7 +13,8 @@
     <title></title>
     <meta charset="UTF-8">
     <script src="../webcomponents-loader.js"></script>
-    <script src="../../web-component-tester/browser.js"></script>
+    <script src="./wct-config.js"></script>
```
sorvell: Cruft, remove this file.

tests/ce-import-upgrade.html
```diff
@@ -13,6 +13,7 @@
     <title></title>
     <meta charset="UTF-8">
     <script src="../webcomponents-loader.js"></script>
+    <script src="./wct-config.js"></script>
```
sorvell: Cruft, remove this file.

tests/ce-import.html
```diff
@@ -13,6 +13,7 @@
     <title></title>
     <meta charset="UTF-8">
     <script src="../webcomponents-loader.js"></script>
+    <script src="./wct-config.js"></script>
```
sorvell: Cruft, remove this file.

webcomponents-loader.js
```diff
@@ -10,15 +10,73 @@
 
 (function() {
   'use strict';
-  // global for (1) existence means `WebComponentsReady` will fire,
-  // (2) WebComponents.ready == true means event has fired.
-  window.WebComponents = window.WebComponents || {};
+
+  var polyfillsLoaded = false;
```
sorvell: Let's add some comments explaining the basic flow here.

webcomponents-loader.js
```diff
@@ -27,48 +85,41 @@
...
-        window.WebComponents.ready = true;
-        document.dispatchEvent(new CustomEvent('WebComponentsReady', {bubbles: true}));
-      });
-    };
-
-    if (document.readyState !== 'loading') {
-      fire();
+    polyfillsLoaded = true;
+    if (document.readyState === 'complete') {
+      ready()
```
sorvell: We can change `ready` to `fireEvent` here since it's not possible to call `waitFor` before this.

README.md
azakus: Done

README.md
azakus: Done

README.md
azakus: Done

tests/ce-import-upgrade-async.html
azakus: Done

tests/ce-import.html
azakus: Done

webcomponents-loader.js
azakus: Done

webcomponents-loader.js
azakus: Done

azakus:
Test failures seem to be related to babel/babel#7506

PR #929, opened 2018-05-08
Reduces the package output

### LINKED ISSUE #928: [v2] NPM package should be stripped
[Browse package here](https://unpkg.com/@webcomponents/webcomponentsjs@2.0.0-beta.2/)

Expected: it does not include tests/, and other extraneous files.

Actual: it includes artifacts in the package.

abdonrd:
Fixes #928.

This reduces the package size.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,5 +55,13 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "files": [
+    "bundles/**/*",
+    "entrypoints/**/*",
+    "custom-elements-es5-adapter.js",
+    "webcomponents-bundle.js",
+    "webcomponents-bundle.js.map",
+    "webcomponents-loader.js"
+  ]
 }


web-padawan:
Consider using `"files"` property instead to specify the included files explicitly: https://docs.npmjs.com/files/package.json#files

abdonrd:
@web-padawan should be like this, right?

```json
  "files": [
    "bundles/**/*",
    "entrypoints/**/*",
    "custom-elements-es5-adapter.js",
    "webcomponents-bundle.js",
    "webcomponents-bundle.js.map",
    "webcomponents-loader.js"
  ]
```

Or am I leaving something?

azakus:
Hmm, is package size a concern?

abdonrd:
@azakus Less is better. And we would save ourselves problems like having to ignore the `gulpfile.js` in the PSK and in the other templates.

abdonrd:
Rebased on master and updated PR branch! :)

keanulee:
+1 on reducing package size. This seems to be what most other npm packages do ([request](https://github.com/request/request/blob/master/package.json), [redux](https://github.com/reactjs/redux/blob/master/package.json), [sw-precache](https://github.com/GoogleChromeLabs/sw-precache/blob/master/package.json), and more).

TimvdLippe:
Thanks!

PR #932, opened 2018-05-08
Implement Symbol polyfill

dfreedm:
Also handle case where loader and bundle are added to the page after
load event, and when they are added between DOMContentLoaded and load
events

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -671,37 +671,6 @@
         "@babel/helper-plugin-utils": "7.0.0-beta.46"
       }
     },
-    "@babel/plugin-transform-modules-commonjs": {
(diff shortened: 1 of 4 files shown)


entrypoints/webcomponents-bundle-index.js
```diff
@@ -31,9 +31,23 @@ if (customElements['polyfillWrapFlushCallback']) {
   });
 }
 
-window.addEventListener('DOMContentLoaded', () => {
+function flushAndFire() {
```
valdrinkoshi: 🚽 & 🔥

valdrinkoshi: love the name :)

PR #944, opened 2018-06-06
:fire: bower

TimvdLippe:
Remove bower as package manager.


TimvdLippe:
Same 2 test failures on the URL polyfill on IE11, other than that working just like master

PR #946, opened 2018-06-07
fix(#942): ie11 compatibility for angular 6

robertfoobar:
This fixes https://github.com/webcomponents/webcomponentsjs/issues/942, i.e. both the stack overflow / endless recursion issue in IE11 as well as the "Not yet implemented errors" thrown by the generated outdated polyfill implementations of google-closure-compiler. 

Unfortunately on my machine, I'm not able to run FF and IE selenium tests. So let's see what CI says. Maybe someone else could verify. 

Also I had to ```npm install -D wd``` in order to get selenium to work. I don't fully understand yet how local testing is supposed to work if that is not included in the package.json.


TimvdLippe:
I had tried this update as well, but the Promise.html test hangs. Closure breaks that polyfill somehow. Therefore, the change is a bit harder than it appears to be.

robertfoobar:
@TimvdLippe Yes, it's harder than I thought. Now that I know that this is not only my local problem, I will try debugging it ... somehow ...

robertfoobar:
I'm really fighting with this. Anybody who has already done wct testing/debugging is warmly welcome to contribute/help ;) e.g.

- How can I get verbose output when running wct on command line?
- How can I run a suite not in command line but in the browser directly?

I'm lacking wct knowledge and need to learn... might take a while.

robertfoobar:
@TimvdLippe Any suggestions/ideas on how to proceed with this PR?

TimvdLippe:
I think @azakus was planning on a deeper analysis in the closure compiler issues, so my advice would be to wait on his work.

robertfoobar:
already fixed

PR #954, opened 2018-06-17
Fix sourcemap urls to not reference into node_modules

### LINKED ISSUE #947: sourcemaps point to incorrect paths
sourcemaps in this repo point to incorrect paths in `sources`.  For example the first source mentioned webcomponents-ce.js.map is `"node_modules/@webcomponents/custom-elements/src/Utilities.js"`.  This is incorrect, it should say `"../component-elements/src/Utilities.js"`.  In theory this could be fixed with the `sourcemaps.mapSources` function.  I'm unable to submit a PR as I cannot get 'npm run build' to work on my system (even without making any change).

TimvdLippe:
Sourcemaps are now hoisted up for easier debugging.

image
Fixes #947


coreyfarrell:
It appears this still has some issues.  I'm unclear what is different now but I'm able to run `gulp`, so I'm doing some testing then I'll give additional feedback.

coreyfarrell:
OK I've looked into this, I think this fix will have to be delayed for gulp-sourcemaps/sources-content/pull/3.  The problem is that we're rewriting the paths before the original source contents are loaded into the sourcemaps.  The referenced PR is to add source to a new module which will allow the sources content to be initialized separately (so we can do it before we rewrite paths).

As is the only reason the build worked reasonably well for you is because you have a folder structure:
```
@webcomponents/webcomponentsjs/  <-- process.cwd()
@webcomponents/custom-elements/
... etc
```
With your patch as-is gulp-sourcemaps is filling the sourcesContent from clones you have in peer folders (`../**`) instead of using the contents from `node_modules/@webcomponents/**`.  Once `@gulp-sourcemaps/sources-content` is implemented we will have to call it before mapSources so we get contents from the correct location.

I have additional changes for gulpfile.js to address remaining issues.  An additional commit is at https://github.com/coreyfarrell/webcomponentsjs/tree/source-map-urls but `npm install` will fail due to the missing dependency (I tested by using `npm pack` on the other module and pointing the local package.json to the tarball).

TimvdLippe:
Oh sorry, this was superseded by #962

azakus:
Ah ok!

PR #955, opened 2018-06-17
Add nocompile to es5-adapter

### LINKED ISSUE #884: Add @nocompile to custom-elements-es5-adapter.js
This will tell closure compiler not to touch the file, including not to translate it from es6 to es5.

TimvdLippe:
Fixes #884

CC @rictic (can't add you as reviewer as you are not in this organisation, but still would like your review :smile: )

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,17 +32,20 @@ function debugify(sourceName, fileName, extraRollupOptions) {
   }
 
   const entry = `./entrypoints/${sourceName}-index.js`;
+  const {output, ...otherExtraRollupOptions} = extraRollupOptions;
+
   const options = {
     input: entry,
     output: {
       format: 'iife',
-      name: 'webcomponentsjs'
+      name: 'webcomponentsjs',
+      ...output
     },
     allowRealFiles: true,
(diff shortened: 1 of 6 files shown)


azakus:
Going to double check that the stanza needs to be at the top or not.

gulpfile.js
```diff
@@ -142,6 +145,9 @@ gulp.task('closurify-bundle', () => {
 
 gulp.task('debugify-ce-es5-adapter', () => {
   const rollupOptions = {
+    output: {
+      intro: '/* @nocompile */'
```
azakus: Needs to be two leading stars: `/** @nocompile */`

gulpfile.js
```diff
@@ -32,17 +32,20 @@ function debugify(sourceName, fileName, extraRollupOptions) {
   }
 
   const entry = `./entrypoints/${sourceName}-index.js`;
+  const {output, ...otherExtraRollupOptions} = extraRollupOptions;
```
TimvdLippe: We can remove all these shenanigans now too, if we add it to the header.

PR #961, opened 2018-06-28
Add @nocompile to ce-es5-adapter

azakus:
v1 port of #955


license-header.txt
```diff
@@ -0,0 +1,9 @@
+/**
+@license @nocompile
```
bicknellr: Is it ok for these to be on the same line?

azakus: sure is!

PR #965, opened 2018-06-29
Provide a baseURI polyfill

### LINKED ISSUE #964: baseURI polyfill is missing in v2
The baseURI polyfill used to be included in HTMLImports, but now that those are gone, the baseURI polyfill is too. This breaks certain elements that relied on it e.g. iron-location for IE

azakus:
This was lost from v1 when HTML Imports was dropped

Fixes #964

### DIFF
--- a/entrypoints/webcomponents-sd-ce-pf-index.js
+++ b/entrypoints/webcomponents-sd-ce-pf-index.js
@@ -23,4 +23,5 @@ import '../node_modules/@webcomponents/shadydom/src/shadydom.js';
 import '../node_modules/@webcomponents/custom-elements/src/custom-elements.js';
 import '../node_modules/@webcomponents/shadycss/entrypoints/scoping-shim.js';
 import '../node_modules/@webcomponents/url/url.js';
+import '../src/baseuri.js';
 import '../src/unresolved.js';
--- a/src/baseuri.js
+++ b/src/baseuri.js
@@ -23,8 +23,8 @@ if (!native_baseURI) {
     get() {
       // this.ownerDocument is `null` for documents
       const doc = this.ownerDocument || this;
-      const base = /** @type {HTMLBaseElement} */ (doc.querySelector('base'));
-      return (base || window.location).href;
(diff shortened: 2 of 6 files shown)


azakus:
@TimvdLippe I couldn't find a reference to an existing test in webcomponentsjs. Maybe that one is in polymer?

src/baseuri.js
```diff
@@ -0,0 +1,32 @@
...
+if (!native_baseURI) {
+  Object.defineProperty(Node.prototype, 'baseURI', {
+    /**
+     * @this {Node}
+     * @return {string}
+     */
+    get() {
+      // this.ownerDocument is `null` for documents
+      const doc = this.ownerDocument || this;
+      const base = /** @type {HTMLBaseElement} */ (doc.querySelector('base'));
```
bicknellr: I think you might want `base[href]` for the selector here.

entrypoints/webcomponents-sd-ce-pf-index.js
```diff
@@ -23,4 +23,5 @@ import '../node_modules/@webcomponents/shadydom/src/shadydom.js';
 import '../node_modules/@webcomponents/custom-elements/src/custom-elements.js';
 import '../node_modules/@webcomponents/shadycss/entrypoints/scoping-shim.js';
 import '../node_modules/@webcomponents/url/url.js';
+import '../src/baseuri.js';
```
bicknellr: Is this not meant to be imported in other entrypoints?

TimvdLippe:
@azakus The `document.baseURI` was previously used in `tests/url.html` and was removed in https://github.com/webcomponents/webcomponentsjs/commit/d10de4d6fe08cab97d7937d2de1a7bfede593b1c

azakus:
Ah yeah, I'll revert that test change

entrypoints/webcomponents-sd-ce-pf-index.js
azakus: No, this is only needed for IE 11

src/baseuri.js
azakus: Good call

PR #966, opened 2018-06-29
make promise polyfill always use MutationObserver timing

azakus:


azakus:
Sorry, I forgot your feedback here. There's a somewhat annoying `setImmediate` issue in IE 10 and 11 that make it unreliable: https://github.com/ReactiveX/rxjs/issues/1163

I'm not 100% sure how to make a small test for it, but we have been hitting it in testing occasionally.

Also, MutationObserver timing is also the most correct for spec microtask timing, which is another hard thing to test for.

PR #977, opened 2018-08-06
Drop es6-promise for promise-polyfill

### LINKED ISSUE #975: Switch 1.0 to taylorhakes promise polyfill
We're currently bundling https://github.com/stefanpenner/es6-promise for our Promise polyfill in webcomponents js 1.0, but https://github.com/taylorhakes/promise-polyfill seems to be smaller, and is what we previously used for our now deprecated https://github.com/PolymerLabs/promise-polyfill fork.

### LINKED ISSUE #978: Promise polyfill flakey on IE11
This is causing race conditions for tests in [LitElement](https://github.com/Polymer/lit-element).

Looks like this is caused by use of `setImmediate`.

azakus:
- Greater compatibility with v2
- Update closure-compiler to same version as v2

Fixes #975
Fixes #978


src/promise.js
```diff
@@ -8,19 +8,45 @@ Code distributed by Google as part of the polymer project is also
...
+  window.Promise = PromisePolyfill;
   // save Promise API
-  ES6Promise.prototype['catch'] = ES6Promise.prototype.catch;
-  ES6Promise.prototype['then'] = ES6Promise.prototype.then;
-  ES6Promise['all'] = ES6Promise.all;
-  ES6Promise['race'] = ES6Promise.race;
-  ES6Promise['resolve'] = ES6Promise.resolve;
-  ES6Promise['reject'] = ES6Promise.reject;
+  /* eslint-disable no-self-assign */
+  // PromisePolyfill.prototype['catch'] = PromisePolyfill.prototype.catch;
```
aomarks: why are these lines commented out?

azakus: I think those are leftover in v2 from figuring out which properties are renamed.

bicknellr: This doesn't look like it should be commented.

src/promise.js
```diff
@@ -8,19 +8,45 @@ Code distributed by Google as part of the polymer project is also
...
+  PromisePolyfill.prototype['then'] = PromisePolyfill.prototype.then;
+  // PromisePolyfill.prototype['finally'] = PromisePolyfill.prototype.finally;
```
bicknellr: `finally` is stage 4 (i.e. will be included in the next spec iteration), should we consider allowing it?

src/promise.js
```diff
@@ -8,19 +8,45 @@ Code distributed by Google as part of the polymer project is also
...
+
+  if (!window.setImmediate) {
+    // approach copied from https://github.com/Polymer/polymer/blob/v3.0.2/lib/utils/async.js
+    const node = document.createTextNode('');
+    const twiddleNode = function twiddleNode() {
+      node.textContent = node.textContent.length > 0 ? '' : 'a';
+    };
+    /** @type {!Array<function():void>} */
+    const callbacks = [];
+    (new MutationObserver(() => {
```
bicknellr: Is it ok to use a `MutationObserver` here? They microtasks but `setImmediate` callbacks are meant to run on new full tasks.

azakus: I think we should probably always use MutationObserver and remove the check for `setImmediate`

src/promise.js
azakus: Figured out only `race` and `all` need to be saved with stringification

src/promise.js
azakus: #978 shows that `setImmediate` is problematic

Issue #987, opened 2018-08-23
v2.1.0 throws when running WCT/Mocha tests with -bundle.js on iOS 11


platosha:
Trying to run a component tests with the `webcomponents-bundle.js` script added to the test suites file on iOS 11.3 or 11.4 Simulator.

Expected result: the tests are running.

Actual result: the error is thrown preventing any tests to run

```
[Error] TypeError: undefined is not an object (evaluating 'Object.getOwnPropertyDescriptor(a,b).enumerable')
	(anonymous function) (webcomponents-bundle.js:69:125)
	filter
	globalProps (mocha.js:1229:3065)
	Runner (mocha.js:1197:1801)
	run (mocha.js:592:4010)
	_runMocha (browser.js:1533)
	_runMocha
	done (browser.js:306)
	wcReady (browser.js:315)
	dispatchEvent
	Lh (webcomponents-bundle.js:250:904)
	(anonymous function) (webcomponents-bundle.js:251:165)
	(anonymous function)
	promiseReactionJob
```

Affected browsers: iOS 11.3 and 11.4 Safari (tried in a Simulator).

Works fine on: desktop Safari and Chrome on macOS, iOS 9.3 Simulator Safari.

I have reproduced the same error (with a slightly shorter stack trace) with `webcomponents-bundle.js` + Mocha alone. Steps:
- Open JSbin: http://jsbin.com/wobaxociya/1/edit?html,console,output
- Open the output URL in Safari in iOS 11 Simulator
- Attach a Web Inspector console to the iOS Safari and reload the tab

web-padawan:
@azakus FYI, regression introduced on #982

platosha:
Here is some more information on the issue: `Mocha.run()` throws on some initial step during `Object.keys(window)`. In particular, there is `showModalDialog` window property that causes this.

On iOS 11.3 and 11.4, `Object.getOwnPropertyNames(window)` lists the `showModalDialog` name among the others, but `Object.getOwnPropertyDescriptor(window, 'showModalDialog')` returns `undefined`.

I would suggest to check for `undefined` descriptor returned by `getOwnPropertyDescriptor`.

web-padawan:
@TimvdLippe could you verify if the 2.1.0 has been build using latest Closure Compiler?

I was working on a PR but noticed the difference making it impossible to fail a test.
Cloned the repo and run `npm i && npm run build` and got the different output:

### Local build: working
```
[...]

PR #988, opened 2018-08-27
Regenerate package-lock.json using npm 6.4.0

web-padawan:
See [#987(comment)](https://github.com/webcomponents/webcomponentsjs/issues/987#issuecomment-416151589)

Also wondering if we can have any check confirming the latest dependencies are used?
Something like `yarn check --integrity` but for npm.


TimvdLippe:
Managed to get this in in #992

PR #1001, opened 2018-09-09
Add HTMLTemplateElement undefined check

zeroliu:
Run into a use case where I have to intentionally remove HTMLTemplateElement.


TimvdLippe:
Closing this per https://github.com/Polymer/lit-html/pull/466#issuecomment-419939869 as this will be polyfilled in `lit-html` instead.

PR #1005, opened 2018-09-14
Fix symbol polyfill

### LINKED ISSUE #998: Version 2.1.0 and 2.1.1 Breaks Reflect Metadata's getMetadataKeys
I'm using SystemJS (version 0.19.46) for the loading of the modules and files in my application, webcomponentsjs for support for native web components, and [reflect-metadata](https://github.com/rbuckton/reflect-metadata) (version 0.1.10) for supporting of metadata. If you need more information about technologies used, please let me know, though it seems pretty cut and dry.

It looks to stem from the overwriting of Object.keys in webcomponentsjs to filter out symbols, which Reflect uses down the line in the call to getMetadataKeys to get the keys which contain metadata (symbols). Here's the code below with the stack trace:

```
system-polyfills.src.js:145 Potentially unhandled rejection [1] TypeError
    at GetIterator (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:823:19)
    at OrdinaryOwnMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:642:24)
    at OrdinaryMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:605:23)
    at OrdinaryMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:609:26)
    at OrdinaryMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:609:26)
    at OrdinaryMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:609:26)
    at Object.getMetadataKeys (https://localhost/jspm_packages/npm/reflect-metadata@0.1.10/Reflect.js:419:16)
    at eval (https://localhost/view-parser.ts!transpiled:182:45)
    at Array.forEach (<anonymous>)
    at ViewParser.parseNode (https://localhost/view-parser.ts!transpiled:180:59)
```

This seems to be an issue mainly with getMetadataKeys, but I imagine it could cause issues elsewhere as well.

### LINKED ISSUE #1000: [BUG] Webcomponentjs breaks Symbol.iterator
After loading `webcomponents-bundle.js` the native `Symbol.iterator` is broken. This bug is introduced in version 2.1.0.

Minimal reproduction: https://codepen.io/ruphin/pen/XPzYBr?editors=1000

ruphin:
This fixes #1000 by loading a vesion of the Symbol polyfill that does not overwrite native implementation.

Fixes #1000 
Fixes #998

### DIFF
--- a/src/symbol.js
+++ b/src/symbol.js
@@ -11,6 +11,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 // import polyfill for Symbol and Object.getOwnPropertySymbols
 import '../node_modules/get-own-property-symbols/build/get-own-property-symbols.max.js';
 
+// Polyfill String.prototype.codePointAt for String iterator
+if (!String.prototype.codePointAt) {
+  const codePointAt = function(i) {
+    let code = this.charCodeAt(i);
+    if (code >= 0xd800 && code <= 0xdbff) {
+        const surr = this.charCodeAt(i + 1);
+        if (surr >= 0xdc00 && surr <= 0xdfff)
+            code = 0x10000 + ((code - 0xd800) << 10) + (surr - 0xdc00);
+    }
+    return code;
+  };
+  Object.defineProperty(String.prototype, 'codePointAt', {
+    value: codePointAt,
+    configurable: true,
+    writable: true
+  });
+}
+
 // overwrite Object.keys to filter out symbols
 Object.keys = function(obj) {
   return Object.getOwnPropertyNames(obj).filter((name) => {
--- a/src/symbol.js
+++ b/src/symbol.js
@@ -11,24 +11,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 // import polyfill for Symbol and Object.getOwnPropertySymbols
 import '../node_modules/get-own-property-symbols/build/get-own-property-symbols.max.js';
 
-// Polyfill String.prototype.codePointAt for String iterator
-if (!String.prototype.codePointAt) {
-  const codePointAt = function(i) {
-    let code = this.charCodeAt(i);
-    if (code >= 0xd800 && code <= 0xdbff) {
-        const surr = this.charCodeAt(i + 1);
-        if (surr >= 0xdc00 && surr <= 0xdfff)
-            code = 0x10000 + ((code - 0xd800) << 10) + (surr - 0xdc00);
-    }
-    return code;
-  };
-  Object.defineProperty(String.prototype, 'codePointAt', {
-    value: codePointAt,
-    configurable: true,
-    writable: true
-  });
-}
-
 // overwrite Object.keys to filter out symbols
 Object.keys = function(obj) {
   return Object.getOwnPropertyNames(obj).filter((name) => {
@@ -39,6 +21,15 @@ Object.keys = function(obj) {
 
 // implement iterators for IE 11
 const iterator = window.Symbol.iterator;
+
+if (!String.prototype[iterator] || !String.prototype.codePointAt) {
+  /** @this {String} */
+  String.prototype[iterator] = function*() {
+    for (let i = 0; i < this.length; i++) {
+      yield this[i];
+    }
+  }
+}
 
 if (!Set.prototype[iterator]) {
   /** @this {Set} */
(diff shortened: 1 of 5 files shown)


TimvdLippe:
Thanks. I have imported the branch and pushed to our repository to let Sauce run on the code.

ruphin:
I verified that this PR builds correctly and that the problem is resolved. However, I cannot run the tests locally, so it's possible that it broke some other things. Let me know if there's any test failures.

ruphin:
I see now that the polyfill does not define Symbol on window, I'll go and see how we can fix that.

ruphin:
Frankly, I think removing the native implementations in this test is the reason why the previous release was broken. Right now there is no test to verify that Symbol and Symbol.iterator still work on browsers that _do_  have native support.

The tests should just check if Symbol and Symbol.iterator work. If the test is run on all platforms, then you will run the tests in environments where native implementations are already present, and in environments where they are not.

TimvdLippe:
So if I understand you correctly, we should simply remove the `delete Symbol` statements and we are good to go?

ruphin:
Exactly.

I think there's still something wrong with the iterator for Strings, the Polyfill implements a String iterator, but it doesn't work for some reason. I'm looking into that now.

ruphin:
The polyfill uses `String.prototype.codePointAt` which is not available on IE11.

We can polyfill that as well, -or- fork the symbol polyfill to remove this dependency.

src/symbol.js
```diff
@@ -24,24 +22,6 @@ Object.keys = function(obj) {
 // implement iterators for IE 11
 const iterator = window.Symbol.iterator;
 
-if (!String.prototype[iterator]) {
```
TimvdLippe: Maybe revert the deletion, but instead change the condition to `!String.prototype[iterator] || !String.prototype.codePointAt`

ruphin: My latest commit adds a miniature polyfill for codePointAt. Your suggestion works too. Let me know which option you prefer.

ruphin: Looks like the polyfilled string iterator is broken for some other reason. I'll try your suggestion

ruphin:
I can't get the tests to pass on IE11, something keeps throwing `Object expected`. I'm going home now to get access to an IE11 machine for debugging.

src/symbol.js
lpellegr: @ruphin I am receiving updates and just saw your last comment about polyfilling String iteration with single index access per character. Is it not going to break apps that iterate on Unicode Strings? For sure, if such an app runs on IE11, it must already include a polyfill, but depending on when webcomponentsjs is loading, it may replace the existing polyfill with a "wrong" one and prevent the existing to load since one is already defined.

What do you think about using the polyfill from Mozilla doc?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#Polyfill

It seems to come from https://github.com/mathiasbynens/String.prototype.codePointAt, which include tests.

ruphin: Thanks! I'll look into it

ruphin: I tried making that work, but for unfortunately the `Symbol.iterator` polyfill for String iteration does not work in IE11, even when using a proper `codePointAt` polyfill.

This change is not changing the existing behaviour of webcomponentsjs, it is only fixing some breakage. The old implementation used the exact same polyfill for String iteration, so it won't break existing apps.

If you have an app that depends on iterator behaviour over Unicode Strings, you can load a 'proper' polyfill for these before loading webcomponentsjs.

ruphin:
It should be working on all platforms now. As far as I can tell CI failed on Safari 10 due to a random Selenium bug.

TimvdLippe:
@ruphin I do not see a commit since 3 days ago. Did you forget to push maybe?

ruphin:
No, I misread the CI build failure three days ago and didn't realise the problem was fixed. If you trigger the CI to run again it will probably work.

TimvdLippe:
Merging this one. Thanks for the work @ruphin !

ruphin:
You're welcome! Any timeframe on the next release? It's awkward to lock to pre-2.1 versions.

TimvdLippe:
I have pinged Dan. We should be able to release today.

azakus:
Published as 2.1.3

Issue #1010, opened 2018-09-19
:host styling does not seem to work on Firefox 62


gotcha:
With the code below, the `Debug` text is displayed as red text on blue background in Chrome or Safari.

However, it is displayed as black test on blue background in Firefox 62.0 on MacOS.

A I missing something ?

```
<html>
    <head>
        <script type="text/javascript" src="https://unpkg.com/@webcomponents/webcomponentsjs@2.1.3/webcomponents-bundle.js"></script>
        <script type="text/javascript" >
        window.onload = function() {
        class MyDebug extends HTMLElement {
        constructor() {
            super()

            let shadowRoot = this.attachShadow({mode: "open"})

            shadowRoot.innerHTML = `
                <style>
                     :host { color: red; }
                     p { background-color: blue }
                </style>
                <p>Debug</p>
            `
          }
        }
        window.customElements.define('my-debug', MyDebug)
        }
        </script>
    </head>
    <body>
        <my-debug></my-debug>
    </body>
</html>
```

web-padawan:
It is not expected to be working that way. See the docs here for usage example:
https://github.com/webcomponents/shadycss#usage

TimvdLippe:
@web-padawan is correct. This is a working in example in Firefox and Chrome: http://jsbin.com/dehimeqipo/edit?html,output

gotcha:
@web-padawan @TimvdLippe Thanks for your answers.

I had read through all this but failed to understand that only :host styling would be broken.

As the rest of the styling of my component was applied, I did not think that the ShadyCSS dance was needed.

In other words, it might be useful to state more explicitely which parts of styling do not work without ShadyCSS.

TimvdLippe:
In your specific example it worked out like this, as `:host` is something specific to shadow dom and older browsers do not understand that. Since the `p` selector is a regular selector, you were seeing the blue background. If you used any other features of shadow dom styling (like `::slotted`), you would have seen the same result. E.g. shadow dom styling not working at all, tainting your whole website.

TimvdLippe:
@gotcha Thank you for the write up! It is very much appreciated to have another point of view of someone new to web components. Feel free to open a PR which clarifies the wording such that you would have instantly understood what to do :smile:

gotcha:
Done ! #1011

PR #1011, opened 2018-09-20
Improve README

gotcha:
See motivation in #1010

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,13 +176,15 @@ polyfills) to be used. We cannot guarantee support for browsers outside of our c
 
 ## Known Issues
 
-  * [ShadowDOM CSS is not encapsulated out of the box](#shadycss)
+  * [Web components styling does not work as described in the web components standard](#shadycss)
   * [Custom element's constructor property is unreliable](#constructor)
   * [Contenteditable elements do not trigger MutationObserver](#contentedit)
   * [ShadyCSS: :host(.zot:not(.bar:nth-child(2))) doesn't work](#nestedparens)
 
-### ShadowDOM CSS is not encapsulated out of the box <a id="shadycss"></a>
-The ShadowDOM polyfill is not able to encapsulate CSS in ShadowDOM out of the box. You need to use specific code from the ShadyCSS library, included with the polyfill. See [ShadyCSS instructions](https://github.com/webcomponents/shadycss).
+### Web components styling does not work as described in the web components standard <a id="shadycss"></a>
+You need to use specific code from the ShadyCSS library (which is already part of webcomponentjs).
+See [ShadyCSS usage](https://github.com/webcomponents/shadycss#usage).
+The reason is that the ShadowDOM polyfill is not able to encapsulate CSS in ShadowDOM out of the box.
 
 ### Custom element's constructor property is unreliable <a id="constructor"></a>
 See [#215](https://github.com/webcomponents/webcomponentsjs/issues/215) for background.
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ polyfills) to be used. We cannot guarantee support for browsers outside of our c
   * [ShadyCSS: :host(.zot:not(.bar:nth-child(2))) doesn't work](#nestedparens)
 
 ### Web components styling does not work as described in the web components standard <a id="shadycss"></a>
-You need to use specific code from the ShadyCSS library (which is already part of webcomponentjs).
+You need to use specific code from the ShadyCSS library (which is already part of webcomponents.js).
 See [ShadyCSS usage](https://github.com/webcomponents/shadycss#usage).
 The reason is that the ShadowDOM polyfill is not able to encapsulate CSS in ShadowDOM out of the box.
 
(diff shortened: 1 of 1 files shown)


gotcha:
- I am afraid a link is still too discrete. And known issues is imho in the same semantic context as browser support.

- using the `style encapsulation` wording demands for the reader to understand what is meant by those words. If he does not know that `:host` selector is related to encapsulation, he might skip over the issue. Also, those words might give the wrong understanding that it is only about CSS rules bleeding in the whole DOM.

web-padawan:
See also #1017

TimvdLippe:
After a discussion with some other devs, we are okay with moving the known issues more above. That at least resolves your first bullet point.

Regarding the second bullet, I have a feeling it comes down to experience. Since I know what the terms are, it is hard to know where someone is confused by. To me, the concept of "style encapsulation" is required to understand shadow dom. Thus if you would understand shadow dom, you would also understand "style encapsulation".

> If he does not know that :host selector is related to encapsulation, he might skip over the issue.

`:host` is not related to style encapsulation. The reason it did not work in your case, was because the browser simply ignored it, as ShadyCSS didn't patch it.

They are thus two distinct issues. The former is related to not preventing styles from leaking into encapsulated dom. The latter is regarding the browser ignoring shadow-dom specific CSS selectors like `:host`.

web-padawan:
@TimvdLippe note the `constructor` issue needs checking, it might be only relevant for v0 polyfill but not sure.

gotcha:
@TimvdLippe I merged after @web-padawan change. Further, I did reword by taking in account your suggestions. Would you review ?

web-padawan:
My gut feeling is that you should be careful about the wording

> does not work as described in the web components standard

That phrase in heading would be stopper for lots of newcomers.
We should highlight that it does not work **automatically** and needs manual tricks.

gotcha:
@web-padawan Good point.

web-padawan:
@gotcha note the comment above regarding the `constructor` issue which is no longer relevant: https://github.com/webcomponents/webcomponentsjs/pull/1011#issuecomment-424757084

README.md
```diff
@@ -174,6 +174,28 @@ for our complete browser support matrix:
...
 polyfills) to be used. We cannot guarantee support for browsers outside of our compatibility matrix.
 
+## Known Issues
+
+  * [Style encapsulation (inline styling in components) does not work out of the box](#shadycss)
+  * [Custom element's constructor property is unreliable](#constructor)
+  * [ShadyCSS: :host(.zot:not(.bar:nth-child(2))) doesn't work](#nestedparens)
+
+### Style encapsulation (inline styling in components) does not work out of the box. <a id="shadycss"></a>
+The ShadowDOM polyfill does not properly support CSS in ShadowDoM out of the box:
```
web-padawan: inconsistent case: `ShadowDoM`, maybe do not need to mention it twice? how about:

> The ShadowDOM polyfill does not properly support scoped CSS in shadow trees out of the box:

README.md
```diff
@@ -174,6 +174,28 @@ for our complete browser support matrix:
...
+ * Any styles inside components have an effect on the whole document (instead of on the component only - the encapsulation is broken).
+ * Any shadow-dom specific selectors (like `:host`) do not work.
+ 
+You can fix those issues by manually calling the `ShadyCSS` APIs. See [ShadyCSS usage](https://github.com/webcomponents/shadycss#usage).
```
TimvdLippe: I still think this section is misleading, as calling the `ShadyCSS` APIs is intended usage behavior when using this polyfill.
What about the following?

```md
To achieve Shadow DOM behavior with regards to CSS, you need to manually call `ShadyCSS` APIs (See [ShadyCSS usage](https://github.com/webcomponents/shadycss#usage) for more details).
If you do not call ShadyCSS, style definitions will have effect on the whole document.
This includes style definitions inside custom elements.
Moreover, `ShadyCSS` will also rewrite (and thus make them work) Shadow DOM-specific selectors like `:host`.
```

TimvdLippe (changes requested):

README.md
gotcha: I understand my wording is incorrect. However the issue that users like me encounter is that , contrary to other polyfills, including ShadowDOM is not enough to make it work according to the web components standard.
And I am trying to make clear what the symptoms of not calling the `ShadyCSS` api are.
Does that seem to be a proper goal ?

TimvdLippe: Yes absolutely agreed that is the goal. My proposed edit is intended to reach that goal as well, is it not?

TimvdLippe: Maybe we should just include a full example in the README here, including how to use `ShadyCSS`, rather than linking to a different doc. Then we can also add a section "CSS selectors not working" and add the solution there. WDYT?

gotcha: That would definitely help. However, I do not volunteer for it as I feel I do not know enough to come up with a meaningful example. Would be happy to review though.

TimvdLippe: I am quite busy atm. I hope to get to this some time next week.

dfreedm (approved):
LGTM

azakus:
I'll remake this PR in the new monorepo: https://github.com/webcomponents/polyfills.

Issue #1014, opened 2018-09-20
possibly deal with Closure warnings


AaronNGray:
```
[18:38:30] Starting 'closurify-ce'...
[18:38:32] gulp-google-closure-compiler: node_modules/@webcomponents/custom-elem
ents/src/CustomElementInternals.js:261: WARNING - actual parameter 1 of Array.pr
ototype.push does not match formal parameter
found   : Element
required: (AlreadyConstructedMarker|HTMLElement|undefined)
    definition.constructionStack.push(element);
                                      ^^^^^^^

node_modules/@webcomponents/custom-elements/src/CustomElementRegistry.js:123: WA
RNING - actual parameter 2 of CustomElementInternals$$module$node_modules$$webco
mponents$custom_elements$src$CustomElementInternals.prototype.setDefinition does
 not match formal parameter
found   : {
  adoptedCallback: (?|undefined),
  attributeChangedCallback: (?|undefined),
  connectedCallback: (?|undefined),
  constructionStack: Array,
  disconnectedCallback: (?|undefined),
  localName: string,
  observedAttributes: (?|undefined)
}
required: {
  adoptedCallback: (Function|null),
  attributeChangedCallback: (Function|null),
  connectedCallback: (Function|null),
  constructionStack: Array<(AlreadyConstructedMarker|HTMLElement)>,
  constructor: Function,
  disconnectedCallback: (Function|null),
  localName: string,
  observedAttributes: Array<string>
}
missing : []
mismatch: [constructor]
    this._internals.setDefinition(localName, definition);
                                             ^^^^^^^^^^

node_modules/@webcomponents/custom-elements/src/CustomElementRegistry.js:124: WA
RNING - actual parameter 1 of Array.prototype.push does not match formal paramet
er
found   : {
  adoptedCallback: (?|undefined),
  attributeChangedCallback: (?|undefined),
  connectedCallback: (?|undefined),
  constructionStack: Array,
  disconnectedCallback: (?|undefined),
  localName: string,
  observedAttributes: (?|undefined)
}
required: (undefined|{
  adoptedCallback: (Function|null),
  attributeChangedCallback: (Function|null),
  connectedCallback: (Function|null),
  constructionStack: Array<(AlreadyConstructedMarker|HTMLElement)>,
  constructor: Function,
  disconnectedCallback: (Function|null),
  localName: string,
  observedAttributes: Array<string>
})
    this._pendingDefinitions.push(definition);
[...]

PR #1019, opened 2018-09-26
Update dependencies to latest version

TimvdLippe:
In preparation for some Closure compiler debugging for #1014


azakus:
We'll do this in the monorepo.

PR #1021, opened 2018-09-27
Add "bare" polyfill bundle

### LINKED ISSUE #1007: Add "bare" polyfill bundle
Currently we ship multiple polyfills with this polyfill. Most notably polyfills for `Symbol`, `URL` and `Promise` (see https://github.com/webcomponents/webcomponentsjs/blob/9871da5b319fd2fe8000c168d6bbb54d905b5a86/entrypoints/webcomponents-sd-ce-pf-index.js) and some stuff regarding `Event` and `Array.from` (https://github.com/webcomponents/webcomponents-platform). Whenever we make an update to these polyfills, some part of our userbase encounters issues.

To reduce the publication risk, it would therefore be better to also publish a polyfill bundle that does not include any other polyfill. It would then be up to the consumer of our polyfill to patch the browser environment themselves.

One option is to introduce a new option `bare`, which would import a `bare` bundle instead on line https://github.com/webcomponents/webcomponentsjs/blob/9871da5b319fd2fe8000c168d6bbb54d905b5a86/webcomponents-loader.js#L139 E.g. change the logic to the following:

```js
polyfills = ['sd-ce-pf' + (window.WebComponents.bare ? '-bare' : '')];
```
With the bare polyfill:

```js
import '../node_modules/@webcomponents/template/template.js';
import '../node_modules/@webcomponents/shadydom/src/shadydom.js';
import '../node_modules/@webcomponents/custom-elements/src/custom-elements.js';
import '../node_modules/@webcomponents/shadycss/entrypoints/scoping-shim.js';
```

TimvdLippe:
Expecting first commit to fail on IE11.

Fixes #1007

### DIFF
--- a/entrypoints/webcomponents-sd-ce-pf-bare-index.js
+++ b/entrypoints/webcomponents-sd-ce-pf-bare-index.js
@@ -1,0 +1,20 @@
+/**
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+'use strict';
+
+/*
+ * Polyfills loaded: HTML Imports, Custom Elements, Shady DOM/Shady CSS, platform polyfills, template
+ * Used in: IE 11
+ */
+
+import '../node_modules/@webcomponents/template/template.js';
+import '../node_modules/@webcomponents/shadydom/src/shadydom.js';
+import '../node_modules/@webcomponents/custom-elements/src/custom-elements.js';
+import '../node_modules/@webcomponents/shadycss/entrypoints/scoping-shim.js';
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -117,6 +117,13 @@ gulp.task('debugify-sd-ce-pf', () => {
   return debugify('webcomponents-sd-ce-pf', null, rollupOptions)
 });
 
+gulp.task('debugify-sd-ce-pf-bare', () => {
+  const rollupOptions = {
+    plugins: [commonjs()]
+  };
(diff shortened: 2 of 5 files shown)


web-padawan:
Any update on this? I would be interested in trying the bare bundle out with Storybook.
Currently it still depends on the v1 polyfill, which could be potentially changed:
https://github.com/storybooks/storybook/blob/d33ac612f9e32dce715ab3c1a1d5ef452ed4488e/app/polymer/src/client/preview/globals.js#L1

TimvdLippe:
I will hopefully have time for this after Chrome Dev Summit (which is in 2 weeks)

web-padawan:
Thanks! Actually, would like to also have a documentation somewhere, about the known edge cases.

As I remember, both Promise and Symbol polyfills are here due to IE11 timing issues with LitElement.
This fact has been mentioned by @sorvell in some issue. 

The absence of the changelog makes us to always remember about those, I even have to keep in mind  some issue numbers, which I occasionally deal with 😆

TimvdLippe:
Yes the Changelog is something we need to work on as well. With CDS it is a bit crazy right now. Hopefully when the dust settles, we can make some real progress in both documentation and serving our power-users with this bare bundle.

TimvdLippe:
Just before CDS we made a plan for this PR. However, some other performance work is having priority. So once that is finished, we will execute the plan and see where we are at.

LarsDenBakker:
Any updates?

LarsDenBakker:
Do you guys have time for look into this now? :)

web-padawan:
@TimvdLippe Can you shed some light on what help is needed to get this merged and released in 2.3.0 ? I would like to help, as this currently blocks me from working on Storybook support for web components and LitElement, storybooks/storybook#4958

I would really like to get rid of the [burden](https://github.com/storybooks/storybook/blob/5bce0de0f723492fe2f95ac54d255cde5c8242e5/app/polymer/src/client/preview/globals.js#L1) they currently have in there:
```js
import '@webcomponents/webcomponentsjs/webcomponents-lite';
import '@webcomponents/webcomponentsjs/custom-elements-es5-adapter';
```

TimvdLippe:
We are still discussing this internally, sorry. This PR resulted in a larger discussion about polyfill loading in general that we want to address first. In the mean time, the open-wc fork might be more suitable for your use case? https://open-wc.org/building/polyfills-loader.html

azakus:
I'm going to redo this PR in the monorepo. Thanks!

PR #1030, opened 2018-10-09
Force Closure to see files as modules by exporting {}

aomarks:
Also remove "use strict" since it's redundant for modules.


azakus:
@aomarks Does this have to do with some ordering issues with Closure?

aomarks:
In a non-module script, `let customElements = window['customElements']` (which is in post-polyfill.js) shadows the global customElements. Since Closure doesn't know a file is a module unless it imports/exports something, Closure throws an error about redeclaring `customElements`.

azakus:
Gotcha, SGTM

azakus:
Can you also upstream this to master?

aomarks:
> Can you also upstream this to master?

Sure, https://github.com/webcomponents/webcomponentsjs/pull/1033

PR #1031, opened 2018-10-09
Set keepFnName in custom-elements-es5-adapter babel config (v1 branch)

aomarks:
Otherwise our HTMLElement definition loses its name.

Effectively we were un-doing the fix from https://github.com/webcomponents/custom-elements/commit/d32780ce115f56abe88a29252f3462a7e94bea61

cc @bicknellr


aomarks:
> LGTM, can you also upstream this to master?

Yep working on it now.

PR #1049, opened 2018-11-02
Patch toString implementation to be correct and strict-mode compatible.

### LINKED ISSUE #1047: Fix Issues with Symbol Polyfill in IE 11
Per https://github.com/es-shims/get-own-property-symbols/issues/13, webcomponentsjs should either look for a different polyfill for Symbol or remove the 'use strict' from the top of the bundled code by setting rollups useStrict config option to false. If going with the first option, core-js would be a good alternative.

kevinpschaaf:
The `get-own-property-symbols` polyfill included in `pf` patches `Object.prototype.toString` that currently relies on loose mode and is incorrect when called against `null`/`undefined` (which mocha does).  When compiled in strict mode (currently done for all bundles), it throws.

There is an issue filed there: https://github.com/es-shims/get-own-property-symbols/issues/16

In the meantime, this patch both makes it compatible with closure's strict-mode compilation and makes the result correct when called against null/undefined.

Fixes #1047

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,53 +10,59 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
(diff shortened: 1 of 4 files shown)


kevinpschaaf:
@TimvdLippe Yup, added tests.

web-padawan:
Any plans to release this fix? 
We would like to try if it resolves problems with bundling polyfill, especially for Angular.

PR #1062, opened 2019-01-10
Fix comment

rictic:


aomarks:
webcomponents-bundle-index.js also mentions HTML imports. Also you might want to instead/also merge this to the html-imports branch instead, since that's what we're syncing too (and will hopefully merge to master).

PR #1063, opened 2019-01-13
Fix loader process flow for no-polyfills cases

### LINKED ISSUE #1056: WebComponents.waitFor() runs twice for async/no-polyfill cases
Hello! According to the README, we are promised that

>The WebComponents.waitFor function takes a callback function as an argument, and will **EVALUATE** that callback **AFTER** the polyfill bundle has been loaded. (emphasis added)

`webcomponents-loader` is designed to handle 4 cases, namely:
1. sync and polyfills needed;
2. sync and no polyfills needed;
3. async and polyfills needed;
4. and async and no polyfills needed.

For case (4), the callback function is actually run twice. This happens for the latest Chrome and FF (v71 and v64 respectively) - FF doesn't require polyfills now that HTMLImports is deprecated. So something like:

```html
<script defer src="https://unpkg.com/@webcomponents/webcomponentsjs@2.2.0/webcomponents-loader.js"></script>

<script type="module">
  WebComponents.waitFor(() => {
    // We want to batch code that requires CE polyfill here so that 
    // polyfillWrapFlushCallback() upgrades all CEs together.
    // But cb actually runs twice for fully speced browsers.
    ...
  });
</script>
```

actually runs the callback twice for compliant browsers.

Here is a super minimal repro: https://jsbin.com/decoreqapo/edit?html,console

Looking deeper into the code, I thought there might be some issues with the implementation. Let me try to be as coherent as I can.

**1. Race condition introduced by runWhenLoadedFns()**

Referencing:

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L77-L91

the `whenLoadedFns` array is cleared **after** its Promise resolves. 

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L81

What probably should happen is it should be cleared **before**, so that if this method is called again before its Promise resolves, evaluation of the functions array wouldn't be repeated. Perhaps this method could be refactored into something like:

```js
  function runWhenLoadedFns() {
    allowUpgrades = false;
    var fnsMap = whenLoadedFns.map(function(fn) {
      return fn instanceof Function ? fn() : fn;
    });
    whenLoadedFns = []; //let GC handle this (ref: https://jsperf.com/array-destroy)
    return Promise.all(fnsMap).then(function() {
      allowUpgrades = true;
      flushFn && flushFn();
    }).catch(function(err) {
      console.error(err);
    });
  }
```

Which leads me to the next issue:

**2. runWhenLoadedFns() is called twice**

TBH I'm not sure if this is intended by design or not. Based on current implementation, for compliant browsers where no polyfills are needed, this block runs.

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L172-L184

The `polyfillsLoaded` flag is asserted right after the browser passes all feature-detection tests. Therefore, referencing `waitFor()` function definition, 

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L95-L103

the flag is `true` and hence `runWhenLoadedFns()` is immediately run when the `WebComponents.waitFor(() => {...})` script is declared. 

However, `runWhenLoadedFns()` is called again via the `ready()` method upon `DOMContentLoaded` event, and before its first call resolves, here:

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L179-L182

This is essentially why the `waitFor()` callback gets evaluated twice for spec-compliant browsers when used asynchronously.

**3. Flow for handling no-polyfill cases seems incorrect**

The current implementation handles no-polyfill cases like this:
* sync - wait for DCL event, run `waitFor()` cbs, then fire WCR.
* async - just fire WCR.

This seems wrong - firstly, async cases would typically have `waitFor()` defined, while sync cases wouldn't, so the implementation got that backwards. Secondly, if we are by design setting `polyfillsLoaded` to true right after the browser passes feature tests over here:

https://github.com/webcomponents/webcomponentsjs/blob/81beb06a05fa3f9c6cc92012b32abb723ad8edb2/webcomponents-loader.js#L173

then there isn't a need to handle anything special except fire the WCR event to indicate that the CEs have been defined.

My apologies that this post got a bit longer than intended - just hoping to seek some clarifications. Thanks!

zerodevx:
Hi @TimvdLippe, 

This fixes #1056. Some notes are:

1. A non-racy implementation of `runWhenLoadedFns()`.

2. For async/no-polyfills cases, we batch all `waitFor()` callbacks and run them in parallel upon the `DCL` event, then fire the `WCR` event after all promises resolve. 

3. For sync/no-polyfills cases, we simply fire the `WCR` event after `DCL`, because we are assured at that point that all CEs will have been defined. 

4. We actually run the same code block in (2) and (3). This works because there cannot possibly be `waitFor` defined in sync cases.

5. Could you check if tests are passing for Edge? I don't have immediate access to a Windows machine..

Thanks so much,
Jason

### DIFF
--- a/tests/runner.html
+++ b/tests/runner.html
@@ -43,5 +43,47 @@
   } catch (e) {} // eslint-disable-line no-empty
 </script>
 <script>
+  // Include `async-loader-no-polyfills` test only for browsers that require no polyfills.
+  // Feature-detection tests are from `webcomponents-loader.js`.
+
+  var polyfills = [];
+  if (!('attachShadow' in Element.prototype && 'getRootNode' in Element.prototype) ||
+    (window.ShadyDOM && window.ShadyDOM.force)) {
+    polyfills.push('sd');
+  }
+  if (!window.customElements || window.customElements.forcePolyfill) {
+    polyfills.push('ce');
+  }
+
+  var needsTemplate = (function() {
+    // no real <template> because no `content` property (IE and older browsers)
+    var t = document.createElement('template');
+    if (!('content' in t)) {
+      return true;
+    }
+    // broken doc fragment (older Edge)
+    if (!(t.content.cloneNode() instanceof DocumentFragment)) {
+      return true;
+    }
+    // broken <template> cloning (Edge up to at least version 17)
+    var t2 = document.createElement('template');
+    t2.content.appendChild(document.createElement('div'));
+    t.content.appendChild(t2);
+    var clone = t.cloneNode(true);
+    return (clone.content.childNodes.length === 0 ||
+        clone.content.firstChild.content.childNodes.length === 0);
+  })();
+
+  // NOTE: any browser that does not have template or ES6 features
+  // must load the full suite of polyfills.
+  if (!window.Promise || !Array.from || !window.URL || !window.Symbol || needsTemplate) {
+    polyfills = ['sd-ce-pf'];
+  }
+
+  if (!polyfills.length) {
+    window.suites.push('async-loader-no-polyfills.html');
+  }
+</script>
+<script>
   WCT.loadSuites(window.suites);
 </script>
--- a/tests/async-loader-no-polyfills.html
+++ b/tests/async-loader-no-polyfills.html
@@ -1,0 +1,72 @@
+<!doctype html>
+<!--
+@license
+Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<html>
+
+<head>
+  <!--
+    This test should be run only for spec-compliant browsers that require no polyfills.
+    When Loader is used async, we expect `WebComponentsReady` event to fire after all
+    waitFor() callbacks are resolved.
+  -->
+  <title>Test loader async/no-polyfills cases</title>
+  <meta charset="UTF-8">
+  <script src="../webcomponents-loader.js" defer></script>
+  <script src="./wct-config.js"></script>
+  <script>
+    WCT.waitForFrameworks = false;
+    window._wctCallback = function() {};
+  </script>
+  <script src="../node_modules/wct-browser-legacy/browser.js"></script>
+  <script>
+    // Define mock load function that takes `duration` to resolve.
+    window.mockLoadCount = 0;
+    window.mockLoad = function(duration) {
+      return new Promise(function(resolve) {
+        window.mockLoadCount++;
+        setTimeout(resolve, duration);
+      });
+    };
+  </script>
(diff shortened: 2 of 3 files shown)


zerodevx:
Sorry for the delay. The test actually passes for Firefox - it works when you call the test runner.html from FF directly, but throws when called from `wct`.. I suspect it's because I set the `WCT.waitForFrameworks = false` flag that causes an issue. I'll try to modify the regression test, and take the opportunity to rebase as well.

zerodevx:
Is this ok?

tests/runner.html
```diff
@@ -42,6 +42,48 @@
     window.suites.push('async-loader-with-modules.html');
   } catch (e) {} // eslint-disable-line no-empty
 </script>
+<script>
+  // Include `async-loader-no-polyfills` test only for browsers that require no polyfills.
+  // Feature-detection tests are from `webcomponents-loader.js`.
```
TimvdLippe: It's unfortunate we are copy-pasting code here, but I can't think of a better solution right now, other than adding another server-roundtrip. That is not something we want to do, so will keep this for now.

TimvdLippe:
Build is running: https://travis-ci.org/webcomponents/webcomponentsjs/builds/482909886

TimvdLippe:
Ugh, 1 test failure on Safari 10. Seems like an actual test failure :(

tests/runner.html
zerodevx: Yup, it's not ideal.. 😞 not a fan of copy-pasting the feature detection tests from the loader too - especially since it should (by right) be kept in sync in future. But the feature detection code is quite standard (and tested) by now, so hopefully there shouldn't be too many changes to it..

zerodevx:
😢

zerodevx:
Can we run Sauce again?

TimvdLippe:
It fixed the timeout issue, so that is progress! Now it is failing with:

```
macos 10.12 safari 10    ✖ tests/async-loader-no-polyfills.html » Loader async/no-polyfills tests » Modules are loaded after WCR fires
  expected undefined to be truthy
    <unknown> at        AssertionError at /components/chai/chai.js:5260:0
    <unknown> at                assert at /components/chai/chai.js:259:0
    <unknown> at             <unknown> at /components/chai/chai.js:581:0
    <unknown> at           addProperty at /components/chai/chai.js:3951:0
    <unknown> at                  isOk at /components/chai/chai.js:1966:0
    <unknown> at   assertModulesLoaded at async-loader-no-polyfills.html:94:0
    <unknown> at             <unknown> at async-loader-no-polyfills.html:104:0
```

zerodevx:
Something strange is going on with Safari 10. Is it possible to trigger Sauce automatically on each commit?

TimvdLippe:
Sauce only runs on branches on our repository, so it can not be automatically triggered on PRs.

zerodevx:
@TimvdLippe I don't expect any test failures now.. 🤞

PR #1075, opened 2019-03-29
Remove old and unused externs.

rictic:
All but `define` don't change the output at all, and `define` only appears to change local variables named 'define'. It looks like these are a holdover from when there was UMD or something in the build.


azakus:
Remaking this PR in the monorepo!

PR #1076, opened 2019-03-29
Modernize closure

rictic:
Builds on top of #1075

* externs are normal srcs
* no need for commonjs anymore
* use the default language_in of "all standards that closure has stable support for"


azakus:
We'll do this in the monorepo for _all_ the polyfills :D

